PR #1, opened 2020-12-04
Standard RPC to replace Convenience and Performance RPCs

MFatihMAR:
[See linked RFC document &rarr;](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/std-rpc-api/text/0000-std-rpc-api.md)

Currently, full document write-up is still WIP but I think we can start having a discussion.
We also have an initial implementation on our internal branch and we'll bring that onto public repo very soon.


text/0000-std-rpc-api.md
```diff
@@ -0,0 +1,433 @@
...
+
+`[ServerRPC]` attribute and matching `...ServerRPC` postfix in the method name are there to make it crystal clear for RPC call sites to know when they are executing an RPC, it will be replicated and executed on the server-side, without necessarily jumping into original RPC method declaration to find out if it was an RPC, if so whether it is a ServerRPC or ClientRPC:
+
+```cs
+Ping(framekey); // Is this an RPC call?
+
+PingRPC(framekey); // Is this a ServerRPC call or ClientRPC call?
+
+PingServerRPC(framekey); // This is clearly a ServerRPC call
+```
```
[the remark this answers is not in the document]
JashanChittesh: I like them being case-insensitive so that they can adapt to different coding styles (I prefer RPC, Rider prefers Rpc and refuses to accept my configuration changes, it seems).

Would it also be possible to allow both, postfixes and prefixes? So, one could write Ping_ServerRPC_ but also _ServerRPC_Ping? At the moment, I feel like the postfix will usually be preferable but there are probably cases where intellisense grouping by method type may be preferable. Postfix is better for "what methods are available, which type of method do I need right now", prefix is better for "what server rpcs does this class offer, ok, I need this specific method".

I think it makes sense to enforce the naming convention - but when it's enforced, it should adapt to different approaches, IMHO.

text/0000-std-rpc-api.md
```diff
@@ -198,6 +201,7 @@ void XyzwServerRpc(int framekey, ServerRpcParams rpcParams = default) { /* ... *
 ```cs
 struct ClientRpcSendParams
 {
+    // who are the target clients?
     ulong[] TargetClientIds;
```
TJHeuvel, line 205: Using an `IEnumerable<ulong>` here would be very convenient for developers, that way i can use a List, Array, NativeArray or even the results of a LINQ query. Accepting an IEnumerable allow us to pass in any collection, as long as it contains identifiers. Additional care has to be taken by the framework not to iterate the targets too often, in the case of a LINQ query lazy evaluation can make this costly.

text/0000-std-rpc-api.md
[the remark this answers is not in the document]
JashanChittesh: One more thing regarding the naming convention: ServerRPC and ClientRPC may be too verbose for some people. So allowing SRPC and CRPC and *maybe* SrvRPC and CliRPC in addition may be appreciated by those. It may not matter much with Ping but sometimes the best name for a method might be DoThatAwesomeThing. In fact, even with Ping, a shorter postfix/prefix may actually improve readability quite a bit.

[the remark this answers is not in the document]
TJHeuvel: Note that something similar is already in place to help with Unity Messages. If the method signature is wrong there is a helpful, yet magic, error message in the Unity Editor.
image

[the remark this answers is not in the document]
TJHeuvel: > The difference with this and other magic-names is that this is actually doing some magic under the hood - we're breaking the general convention of, "when you call a function, it will be invoked synchronously before returning control to the parent function". This is going to make comprehension a lot more difficult for users if these functions look like every other function at the call site. Without drilling down, you have no way of knowing that the function you're calling is actually an RPC that will be invoked asynchronously.
> I wouldn't say that the naming suffix requirement is an elegant solution, but it does at least improve call site readability. Now the user can build some expectations when reading code, "ok this behavior can only happen when I see this suffix". Even though it's a little awkward for the end user, I think we end up saving them a lot of pain if we do this.

This i think is very important to mention, and is something that trips up new users already. A few days ago on the Discord a user was confused about the fact that it doesnt execute immediately.

text/0000-std-rpc-api.md
```diff
@@ -0,0 +1,397 @@
...
+We're looking at **up to 5.24x better performance** (compared to convenience API vs standard API receive performance).
+
+Beyond that, on an internal project, we saw roughly **%80 less boilerplate code** compared to performance RPCs and standard RPCs which is also a huge win.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Multiplayer framework provides 2 main network constructs ([ServerRpc](#serverrpc) and [ClientRpc](#clientrpc)) to execute logic on either server-side or client-side. This concept often called as [Remote Procedure Call (RPC)](https://en.wikipedia.org/wiki/Remote_procedure_call) and has wide adoption across the industry.
+
+## RPC Methods
```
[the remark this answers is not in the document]
TJHeuvel: I cant find the subtitle about response.

If this is implemented, i'd really like to see a way to wait for the result. Unity itself uses Coroutines a lot, but my personal preference is an async task based API.

text/0000-std-rpc-api.md
```diff
@@ -0,0 +1,393 @@
...
+```
+
+Developer can invoke a `ClientRpc` by making a direct function call with parameters:
+
+```cs
+void Update()
+{
+    if (Input.GetKeyDown(KeyCode.P))
+    {
+        PongClientRpc(Time.frameCount); // Server -> Client
```
TJHeuvel, line 108: What happens when i call PongClientRPC on a Client?

I would understand that it would throw an error, but coming from a Photon PUN background, and being used to relay, it would be very convenient that this would become an automatic-relay-RPC. The server would recieve the message, and immediately relay it to the clients, optionally defined in the ClientSendParams argument.

text/0000-std-rpc-api.md
```diff
@@ -0,0 +1,393 @@
...
+
+### ServerRpc
+
+A `ServerRpc` can be invoked by a client to be executed on the server.
+
+Developer can declare a `ServerRpc` by marking a method with `[ServerRpc]` attribute and making sure to have `ServerRpc` suffix in the method name.
+
+```cs
+[ServerRpc]
+void PingServerRpc(int framekey) { /* ... */ }
```
TJHeuvel, line 54: How can this setup work with a, as MLAPI describes it, [multiple project setup](https://mlapi.network/wiki/messaging-system/)? E.g. my server is a completely different Unity project, the `server` methods dont exist in the client project and vice-versa. In MLAPI i can invoke methods by string parameter.

Maybe a similar scenario can be made in the same project, where in order to separate concerns i split Server and Client up in different scripts. The Server adds the `ServerManageGameState` NetworkedBehaviour, the client adds the `ClientManageGameState` NetworkedBehaviour. 

How can these two scripts send RPCs to eachother?

text/0000-std-rpc-api.md
```diff
@@ -0,0 +1,397 @@
...
+[ServerRpc]
+void AbcdServerRpc(int framekey, ServerRpcSendReceiveParams sendReceiveParams = default) { /* ... */ }
+```
+
+### ClientRpc Params
+
+```cs
+struct ClientRpcSendParams
+{
+	ulong[] TargetClientIds;
```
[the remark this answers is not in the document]
TJHeuvel: https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/pull/1#discussion_r539214932

I've asked for this to be an IEnumerable instead, that way you dont have to allocate memory at all in many cases.

text/0000-std-rpc-api.md
TJHeuvel, line 108: When is this RPC sent?

Is there an built-in mechanism for doing rate-limiting, or linking it to a tick rate?  If not, what happens when i create my own system for that, and invoke the RPC as a delegate?


   ```

class RPCTest
{
    void OnPlayerConnected()
    {
        Scheduler.ScheduleRPC(Scheduler.CurrentServerTick + 5, PongClientRPC, "Hello World!");
    }

    [ClientRPC]
    void PongClientRPC(string message)
    {

    }

}

public static class Scheduler
{
    public static int CurrentServerTick { get; } //TODO: Actual tick, or time
    public static async void ScheduleRPC<T>(int atTick, Action<T> rpc, T argument)
    {
        while (CurrentServerTick != atTick) await Task.Yield();

        rpc(argument);
    }

}
```

Does this invoke the RPC?

Comment (TJHeuvel) [inline text/0000-std-rpc-api.md:108]: When is this RPC sent?

Is there an built-in mechanism for doing rate-limiting, or linking it to a tick rate?  If not, what happens when i create my own system for that, and invoke the RPC as a delegate?


   ```

class RPCTest
{
    void OnPlayerConnected()
    {
        Scheduler.ScheduleRPC(Scheduler.CurrentServerTick + 5, PongClientRPC, "Hello World!");
    }

    [ClientRPC]
    void PongClientRPC(string message)
    {

    }

}

public static class Scheduler
{
    public static int CurrentServerTick { get; } //TODO: Actual tick, or time
    public static async void ScheduleRPC<T>(int atTick, Action<T> rpc, T argument)
    {
        while (CurrentServerTick != atTick) await Task.Yield();

        rpc(argument);
    }

}
```

Does this invoke the RPC?

text/0000-std-rpc-api.md
```diff
@@ -0,0 +1,397 @@
...
+PongRpc(framekey); // Is this a ServerRpc call or ClientRpc call?
+
+PongClientRpc(framekey); // This is clearly a ClientRpc call
+```
+
+### Reliability
+
+RPCs are reliable by default which means they are guaranteed to be executed on the remote side. However, sometimes developer might want to opt-out reliability, which is often the case for non-critical events such as particle effects, sounds effects etc.
+
+Reliability configuration can be specified for both `ServerRpc` and `ClientRpc` methods at compile time:
```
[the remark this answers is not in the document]
TJHeuvel: Why is this in the method name, instead of the RPCParams?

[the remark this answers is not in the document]
MFatihMAR: It's not in the method name, it's in the attribute: `[ServerRpc(IsReliable = false)]`

text/0000-std-rpc-api.md
TJHeuvel: A built in mechanism could use the RPCSendParams?

text/0000-std-rpc-api.md
```diff
@@ -0,0 +1,393 @@
...
+|ServerRpc Network Call|✅|❌|✅|
+|ServerRpc Direct Call|❌|❌|❌|
+|ClientRpc Network Send|✅|❌|✅|
+|ClientRpc Network Call|❌|✅|✅|
+|ClientRpc Direct Call|❌|❌|❌|
+
+## RPC Params
+
+// todo: descriptions needed below
+
```
JashanChittesh, line 172: Unity's old legacy networking (and consequently, PUN, which is mentioned in the Prior Work section in the RFC) has [NetworkMessageInfo](https://docs.unity3d.com/560/Documentation/ScriptReference/NetworkMessageInfo.html) ([PhotonMessageInfo](https://doc-api.photonengine.com/en/pun/v2/struct_photon_1_1_pun_1_1_photon_message_info.html) in the case of PUN). Both include timestamps.

With a reliable NetworkTime (like [Network.time](https://docs.unity3d.com/560/Documentation/ScriptReference/Network-time.html)), this makes it comparatively easy to estimate the real position based on position when the RPC was sent and RPC transition time (and velocity). It's also useful for sanity checks on whether a given RPC is (still) valid.

This is useful for both, ClientRPC and ServerRPC and is currently missing in both. It depends on having reliable network time, of course.

text/0000-std-rpc-api.md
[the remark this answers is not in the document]
TJHeuvel: Why do i both need an attribute and the naming convention to match? Isnt the attribute redundant?

text/0000-std-rpc-api.md
MFatihMAR: RPCs won't be replicated immediately but rather queued to be sent.
We will have our own network loop that ticks very early in the frame and very late at the end.
At the end of a frame, we will do things and send RPCs down to sockets.

text/0000-std-rpc-api.md
MFatihMAR: I think it'd make sense to have more context info in the `XXXReceiveParams`. This RFC only a future-proof way to add extra info later without breaking existing RPC API. One of those upcoming additional fields might well be network time, timestamp, frame number etc.

text/0000-std-rpc-api.md
```diff
@@ -0,0 +1,393 @@
...
+```
+
+Developer can invoke a `ServerRpc` by making a direct function call with parameters:
+
+```cs
+void Update()
+{
+    if (Input.GetKeyDown(KeyCode.P))
+    {
+        PingServerRpc(Time.frameCount); // Client -> Server
```
TJHeuvel, line 64: What happens when i connect to a server which has a different network signature? E.g. my client uses an older version where this was an `int`, the server is up to date and this uses a `string`. 

How does the server handle this case? Is the entire RPC dropped?

text/0000-std-rpc-api.md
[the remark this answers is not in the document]
MFatihMAR: > `[ServerRpc]` attribute and matching `...ServerRpc` suffix in the method name are there to make it crystal clear for RPC call sites to know when they are executing an RPC, it will be replicated and executed on the server-side, without necessarily jumping into original RPC method declaration to find out if it was an RPC, if so whether it is a ServerRpc or ClientRpc:
> 
> ```cs
> Ping(framekey); // Is this an RPC call?
> 
> PingRpc(framekey); // Is this a ServerRpc call or ClientRpc call?
> 
> PingServerRpc(framekey); // This is clearly a ServerRpc call
> ```

text/0000-std-rpc-api.md
```diff
@@ -0,0 +1,393 @@
+- Feature Name: `std-rpc-api`
+- Start Date: 2020-11-02
+- RFC PR: [Unity-Technologies/com.unity.multiplayer.rfcs#0000](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/pull/0000)
+- Issue: [Unity-Technologies/com.unity.multiplayer#0000](https://github.com/Unity-Technologies/com.unity.multiplayer/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Existing MLAPI [**Convenience** and **Performance** RPCs](https://mlapi.network/wiki/messaging-system/) are sub-optimal in terms of required boilerplate code, API design and runtime performance.
```
TJHeuvel, line 9: The Performance API allows us to read/write to a stream, that is sometimes quite convenient. Is there some equivalent, e.g. can i make a method:
```

[ClientRPC]
void RecieveDataClientRPC(DataStreamReader stream)
{
 if(stream.Length == 1)
  int data = stream.ReadInt();
 else if(stream.Length == 2) {
  int data1 = stream.ReadInt();
  int data2 = stream.ReadInt();
} 
}
```

NoelStephensUnity: I believe that the INetworkSerializable interface provides this ability:
```
public interface INetworkSerializable
{
        void NetworkRead(BitReader reader);
        void NetworkWrite(BitWriter writer);
}
```
Since the standard RPC framework yields better performance than both the original convenience and performance RPC, the only real use for that level of access would be for scenarios where you want to send specific sets of data or perhaps you want to handle things yourself.  Using a INetworkSerializable derived class/structure as an RPC parameter yields similar functionality to that of the traditional performance RPC when it comes to being able to read and/or write into a stream.

text/0000-std-rpc-api.md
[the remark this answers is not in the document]
xCocoDev: Hey @MFatihMAR, does Reliable RPC would also mean that the ORDER in which they are sent is kept/ensured?

text/0000-std-rpc-api.md
JashanChittesh: I'd consider RPC-stuff a "distributed API", kind of like REST. For games, **usually** you'll have enough control to make sure clients and servers are on the same version. Ever had a game (or app) tell you "you need to update to play" or "you need to update to play _online_"? That's why.

If you really need to support different versions talking to each other, I'd first check if you **really** need to support different versions. By supporting this, you're making an already extremely complex thing even more complex (with lots of really nasty bugs you might run into that will be a nightmare to isolate, reproduce and fix).

So most likely, if you've done it right, the server will kick the client if the versions don't match, problem solved. This, btw, also helps with cracked versions (that often are outdated). It can also help, to a certain degree, with clients that would cheat.

If you still really need to support different versions working together: Include version numbers in your method names. In all of them. And tag old method names with Obsolete to make builds fail when still accessing them. In that case, it's kind of like versioning for REST-APIs, basically "URI versioning", like here: https://restfulapi.net/versioning/

But most likely, you will regret having made that choice that after about a week ;-)

MFatihMAR: cross-build and cross-version compatibility is something we need to discuss separately. having said that, MLAPI doesn't support cross-build and/or cross-version compatibility today.

text/0000-std-rpc-api.md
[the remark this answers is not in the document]
MFatihMAR: Fantastic question @xCocoDev! I think we need to have a bit of a discussion here.

I propose having them ordered on the `NetworkObject` but not globally ordered.
I mean, RPCs fired on the same `NetworkObject` will arrive in order but RPCs called on different `NetworkObject`s might end up being called out of order (still in order internally but not in order globally).

I will also bring that up in one of our meetings internally to build a consensus here inside our team and inline our proposal on the markdown document side as well.

text/0000-std-rpc-api.md
MFatihMAR: Please see "Execution Table" section.

text/0000-std-rpc-api.md
MFatihMAR: There is no cross-binary (and also no cross-version) compatibility in MLAPI today. Also an RPC can't map to another object. What you can do is though, have a `NetworkBehaviour` full of `ClientRpc`s and `ServerRpc`s, then call `ClientRpc`s inside `ServerRpc`s and that'd just work fine.

text/0000-std-rpc-api.md
MFatihMAR: For more details, see "Execution Table" section.

text/0000-std-rpc-api.md
MFatihMAR: This RFC only covers RPCs but not necessarily all of the replication mechanisms on `NetworkBehaviour`s. We can address manual serialization and replication under another RFC (potentially called `NetworkBehaviour` replication).

And also as Noel said, you could extend serialization for your own types via `INetworkSerializable` interface.

text/0000-std-rpc-api.md
JashanChittesh: Cross-build compatibility is not supported? To me, that would mean that if you do a server-build, a build for Windows and a build for Android, those could not talk to each other. That doesn't sound plausible, so I assume you mean something else with "cross-build"? But then, MLAPI would also support "cross-version" as long as people don't mess with the interfaces / method signatures.

Ah ... maybe that's it: So MLAPI does not support two builds talking to each other, if there was any change in any of the methods that belong to the RPC-contract? That would make total sense to me.

Is that documented somewhere? It's a limitation that makes total sense - but it's also a limitation that people really need to be aware of. My understanding would be that any change to any RPC would break compatibility but Custom Messages should be fine.

text/0000-std-rpc-api.md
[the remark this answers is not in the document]
TJHeuvel: The more i think about it, the less i like a C# framework forcing naming conventions on us. Several reasons:

1. Although its a fine guideline, others can be proposed that are equally good/bad. Personally i started prefixing methods with `client` or `server`, that made most sense for me. I prefix because otherwise you get methods that are called `validateClientServerRpc`, vs `serverValidateClient`. I find the latter is clearer.
2. Some developers have code bases in their native language, where this wouldnt fit well.
3. It might make migrating from another networking stack more difficult.
4. Within C# this is very uncommon, besides the magical monobehaviour callback methods Unity has there is as far as i am aware no other mainstream framework that does it. There is no prior-art. 
5. There is no technical reason, the attribute is enough. A strong recommendation achieves nearly the same.

TwoTenPvP:
I still think direct invokes are a bad idea.

The proposal here is to change how C# works.

```csharp
MyRPC();
```
I fully expect this to run the MyRPC method. That's how C# works.

What is the advantage of replacing direct invokes instead of having a invoke method like the MLAPI has today? I think this is much cleaner, it also works much better with arguments.

```csharp
InvokeOnNetwork(MyRPC, myOptions);
```

What is the actual advantage here to not use a Invoke method? The disadvantage is confusion and limitation for no reason. Like reusal of methods for both local and remote invokes.

text/0000-std-rpc-api.md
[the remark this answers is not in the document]
MFatihMAR: @TJHeuvel — I think that's fair, enforcing name convention by hard compile errors seems a little bit harsh. Maybe we can stick with compile/build warnings instead of hard-errors to attract more attention from the developer.
_(I think I'm changing my mind around too here)._

Microsoft does something similar with Async APi where you're overwhelmingly encouraged to put `...Async` suffix to your async task methods but you're not forced to. Usually you see a warning message popping up on that method signature and method call sites advising you to use `...Async` suffix but not forcing it hard.

It all comes down to "is tooling will be enough?" and it's also fair to say Unity Diagnostics Messages and Roslyn Analzyers will be covering that part thoroughly.

(I can see debug time spent figuring out what that method didn't replicate or vice-versa even today but after all, you can only enforce so much)

LukeStampfli:
Elaborating a bit on @TwoTenPvPs comment to use invoke instead of direct function call with an alternative proposal:

The idea is simple, retain the old interface of convenience RPCs but get all the performance and serialization benefits of the Standard RPC proposed in this RFC.

## guide-level explanation
RPCs are invoked with 'InvokeServerRPC' etc. parameters are passed into the invoke function. E. g.
```csharp
[ServerRpc]
void PingServerRpc(int framekey) { /* ... */ }

void Update()
{
    if (Input.GetKeyDown(KeyCode.P))
    {
        InvokeServerRpc(PingServerRpc, Time.frameCount); // Client -> Server
    }
}
```

ClientRpcReceiveParams can be added to an RPC function and will be automatically filled in on the receiving side. E. g.
```csharp
[ServerRpc]
void PingServerRpc(int framekey, ServerRpcReceiveParams params = default ) { /* ... */ }
```

## Reference-level explanation
### Pros
- No breaking changes for users which use the current 'convenience RPCs'
- IL post processing could be deactivated and the code would still work.
- The interface for invoking RPCs is reguluar c# and not magic
- Works for multi project setups (separate client and server binary)
### Drawbacks
- If we support deactivating the weaver we might end up with overhead to implement features twice once for ILPP and once for regular c# without ILPP.
- We have to keep the hundreds of template functions in NetworkedBehaviour.BoxedServerRPCs etc.

## Open Questions
- If we go this route do we support RPCs without ILPP?
- What are we loosing by using this API over the one proposed in the initial RFC

TwoTenPvP:
> Elaborating a bit on @TwoTenPvPs comment to use invoke instead of direct function call with an alternative proposal:
> 
> The idea is simple, retain the old interface of convenience RPCs but get all the performance and serialization benefits of the Standard RPC proposed in this RFC.
> ## guide-level explanation
> 
> RPCs are invoked with 'InvokeServerRPC' etc. parameters are passed into the invoke function. E. g.
> 
> ```cs
> [ServerRpc]
> void PingServerRpc(int framekey) { /* ... */ }
> 
> void Update()
> {
>     if (Input.GetKeyDown(KeyCode.P))
>     {
>         InvokeServerRpc(PingServerRpc, Time.frameCount); // Client -> Server
>     }
> }
> ```
> 
> ClientRpcReceiveParams can be added to an RPC function and will be automatically filled in on the receiving side. E. g.
> 
> ```cs
> [ServerRpc]
> void PingServerRpc(int framekey, ServerRpcReceiveParams params = default ) { /* ... */ }
> ```
> 
> ## Reference-level explanation
> ### Pros
> 
>     * No breaking changes for users which use the current 'convenience RPCs'
> 
>     * IL post processing could be deactivated and the code would still work.
> 
>     * The interface for invoking RPCs is reguluar c# and not magic
> 
>     * Works for multi project setups (separate client and server binary)
> 
> 
> ### Drawbacks
> 
>     * If we support deactivating the weaver we might end up with overhead to implement features twice once for ILPP and once for regular c# without ILPP.
> 
>     * We have to keep the hundreds of template functions in NetworkedBehaviour.BoxedServerRPCs etc.
> 
> 
> ## Open Questions
> 
>     * If we go this route do we support RPCs without ILPP?
> 
>     * What are we loosing by using this API over the one proposed in the initial RFC

Well explained.

I think there is no advantage to allowing you to run without ILPP. Just more complexity.
It also removes one of the drawbacks.

text/0000-std-rpc-api.md
MFatihMAR: Yep, this:

> Ah ... maybe that's it: So MLAPI does not support two builds talking to each other, if there was any change in any of the methods that belong to the RPC-contract? That would make total sense to me.

I will try to inline this conversation too.

TwoTenPvP: The MLAPI does support different builds talking to each other. They just need to have the same RPC name.

text/0000-std-rpc-api.md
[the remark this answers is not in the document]
MFatihMAR: inlined:
> Reliable RPCs will be received on the remote end in the same order as they are fired but this in-order guarantee only applied to RPCs on the same `NetworkObject`. Different `NetworkObject`s might have reliable RPCs called but executed out of order compared to each other. To put more simply, in-order reliable RPC execution is guaranteed per `NetworkObject` basis only.

MFatihMAR:
@TwoTenPvP

> It also removes one of the drawbacks.

Can you elaborate a little more, specifically?

---

@LukeStampfli 

> - IL post processing could be deactivated and the code would still work.

If you strip out the framework, how it will continue working? I wonder how do you define "work". Do you mean if it compiles it works or if it's doing what's expected to happen then it works?

> - The interface for invoking RPCs is regular c# and not magic

I also wonder how do you define "regular C#". Does it mean just a C# console app? Looking from another perspective you could even argue async-await syntax is also magic (I'd rather say it's syntactic sugar). Another example might be Asp.Net and EntityFramework. If you strip out the whole Asp.Net and/or EF frameworks, you are not going to change your code but it will be broken. Asp.Net does dependency injection for you, also EF does lots of "magic" and how is that standard C# but Unity is not? RPC syntax is a new network construct we're bringing into C# in Unity, it's more of a "syntactic sugar" but you can also call it "magic" but I can tell lots of Microsoft projects are using "magic". If we are saying console app C# is the standard C#, I'd disagree. If we are saying only Microsoft branded syntactic sugars are allowed in C#, I'd also disagree.

> - Works for multi project setups (separate client and server binary). The new proposed API doesn't

Can you elaborate more?

> - Same performance as currently proposed standard RPC API.

Can I see some numbers and/or can you elaborate how that'd be achieved/implemented?

> - Send and receive params are clearly split. Instead of having a struct which contains both send and receive params the invoke function takes the send params and the receive function the receive params only. This makes the API cleaner.

We can still do this by allowing either one of these on a RPC method signature: `ClientRpcSendParams`, `ClientRpcRecvParams`, `ServerRpcSendParams`, `ServerRpcRecvParams`.
Packing them into one is basically saying we have only 2 options: with and without params. We already discussed that earlier up there. Main concern was making it too verbose so we wanted to only inject/consume one param rather than 3 (send, recv, sendrecv).

---

I would even go one step further and invite you to have a look at prior art (aka how other people do this). Unreal RPCs might be a good start but I can tell you that direct function call without extra framework utilities in the middle is the most loved and embraced option in the industry. There is fundamentally no need for extra steps and wrappings to make an RPC call, we bring that into Unity/C# as a syntactic sugar — or course it wouldn't work without the framework. Majority of Unity/C# wouldn't work without Unity runtime and frameworks. Lots of Microsoft branded products wouldn't work without their own framework environment. What is the difference here?

I'm happy to consider this discussion in a separate RPC API RFC as a successor but so far, it all sounds like personal preferences to me.

TwoTenPvP:
> Can you elaborate a little more, specifically?
The drawback ``If we support deactivating the weaver we might end up with overhead to implement features twice once for ILPP and once for regular c# without ILPP.``
> I also wonder how do you define "regular C#". Does it mean just a C# console app? Looking from another perspective you could even argue async-await syntax is also magic (I'd rather say it's syntactic sugar). Another example might be Asp.Net and EntityFramework. If you strip out the whole Asp.Net and/or EF frameworks, you are not going to change your code but it will be broken. Asp.Net does dependency injection for you, also EF does lots of "magic" and how is that standard C# but Unity is not? RPC syntax is a new network construct we're bringing into C# in Unity, it's more of a "syntactic sugar" but you can also call it "magic" but I can tell lots of Microsoft projects are using "magic". If we are saying console app C# is the standard C#, I'd disagree. If we are saying only Microsoft branded syntactic sugars are allowed in C#, I'd also disagree.

C# is a language, with a language spec. In C#, the following syntax is used to jump to the code in said method body:
```csharp
Hello()
void Hello(){}
```
We are changing that fact. And why are we doing that? What are the advantages? This has led to confusion in the past in other libs, it's one thing people have valued with the MLAPI, it was pure C#. Easy to see what's going on.

> Can I see some numbers and/or can you elaborate how that'd be achieved/implemented?
We are not chancing anything, performance will be identical. We are just adding a dummy Invoke method which is replaced with whatever the native call would be replaced with. 

>  but I can tell you that direct function call without extra framework utilities in the middle is the most loved and embraced option in the industry

Where is the source for that? People like the MLAPI for the exact opposite reason.



I will restate my question. Where is the advantage in changing C#? What are we gaining?

LukeStampfli:
> If you strip out the framework, how it will continue working? I wonder how do you define "work". Do you mean if it compiles it works or if it's doing what's expected to happen then it works?

What I'm saying is that we would have the option to deactivate the weaver and just use the currently existing code in MLAPI to do run the RPCs. I'm not sure whether that's a smart option to do, probably not but we have the option to make it work without ILPP. The API in the RFC proposal can strictly not work without IL2PP/ source generators its 100% impossible.

> I also wonder how do you define "regular C#".

A function call calls a method. MyServerRpc() generates IL code which makes a regular function call. The API which we are proposing breaks c# semantics. EF framework does exactly what I'm proposing. It uses generics and weaves behind the box, the API to access it is regular c#.

I'm a fan of self documenting code. If I write InvokeServerRpc() it is 100% clear what that code does. It sends a message over the network to call an RPC. MyServerRpc() on the other hand isn't. Someone needs a degree in 'MLAPI' to understand what that call does. How does it behave on a host does it also go over the network pipeline (yes). What if I want to call my RPC function directly etc. There are so many questions which need to be documented.

> Works for multi project setups (separate client and server binary). The new proposed API doesn't

With my proposed syntax we can do InvokeServerRpc("MyServerRpc") without the need for having the method definition in the same assembly. This allows to very easily split server and client assemblies/projects.

> Can I see some numbers and/or can you elaborate how that'd be achieved/implemented?

I can go into implementation details:
`InvokeServerRpc(Foo)` results in IL like:
```
ldarg 0
ldarg 0
create delegate
for each param {
    ldarg
} // load params onto the stack
call virtual InvokeServerRpc
```

We manipulate this in the following ways:
- strip the delegate creation
- create bitwriter and call BeginWrite before loading params on the stack
- write method hash onto the stack (we can figure this out based on the delegate/string instruction)
- use existing IL to load params onto the stack
- foreach param call the right writer function to write the param (ld arg the writer, for the value param we use what already exists on the stack.
- replace virtual call to InvokeServerRpc with EndWrite

text/0000-std-rpc-api.md
JashanChittesh: > This is all because of the deterministic hashing of RPC names.

I think this is a very important thing to keep. @MFatihMAR The mapping of RPCs being non-deterministic (i.e. one change could make the whole thing break) seems like an implementation choice that will be easy to regret. It does seem to me that this would be an implementation-level issue (not a design issue) but I don't know enough about the "magic" that would wire up RPCs with IDs.

JashanChittesh:
> Where is the advantage in changing C#? What are we gaining?

More readable code? If the naming convention with the postfixes (or prefixes) is kept, the method name has all the information you need to understand what happens. With Invoke, you have more boiler-plate code and an unusual syntax for calling methods.

`someInstance.InvokeClientRpc(someInstance.MyClientRPC, param1, param2, param3);`

vs.

`someInstance.MyClientRPC(param1, param2, param3);`

IMHO, that's a very significant advantage. One (also fairly strong) argument **against** this is that it's sometimes nice to be able to call RPCs directly. But in my experience, usually it's more elegant to have the method that can be called directly as its own method, and any networked related stuff in the RPC-method(s), and then calling the regular method from the RPC.

TwoTenPvP:
> More readable code? If the naming convention with the postfixes (or prefixes) is kept, the method name has all the information you need to understand what happens. With Invoke, you have more boiler-plate code and an unusual syntax for calling methods.

That is the point though. You are not calling the method. You are sending a packet for something to be invoked on the remote. It's misleading. We are using the C# syntax for calling a method for something other than calling the method.

I have never seen anyone complain about having to specify ``InvokeOnNetwork(MyRpc)``. It makes sense.

text/0000-std-rpc-api.md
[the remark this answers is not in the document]
gIsForGravity: @TJHeuvel, I don't think it really makes sense to use async tasks for this, since messages are being polled during frames. Using async for this would be kind of pointless since it wouldn't receive the message any quicker. If you really need to, spin off a task during the coroutine.

gIsForGravity:
> What is the actual advantage here to not use a Invoke method? The disadvantage is confusion and limitation for no reason.

There are two advantages. One, code looks a lot cleaner. Two, new people don't get confused when they call their function normally and it runs locally.

However I am not saying that not using an Invoke method is good here, just pointing out the advantages. I'm sure new users would get confused when their RPCs aren't running synchronously when they call them normally, and the code really isn't *that* much messier with the Invokes. So I would agree that it's probably better to keep the invokes.

JashanChittesh:
> You are not calling the method. You are sending a packet for something to be invoked on the remote. It's misleading. We are using the C# syntax for calling a method for something other than calling the method.

Basically, our method call in that case is equivalent with calling `DoTheCoolThingServerRPC(3, 4)`, when that method is:

`public void DoTheCoolThingServerRPC(int param1, int param2) {
    InvokeOnServer(DoTheCoolThingActualServerRPC, param1, param2);
}`

The only difference is that people don't have to write these method declarations for each and every RPC.

text/0000-std-rpc-api.md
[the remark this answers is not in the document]
becksebenius-unity: One alternative to this is to provide a project level MLAPI setting for the naming convention of methods. Two regex fields - "ServerRpcValidation" and "ClientRpcValidation". An empty string would indicate no validation, and otherwise the regex would be used in the compilation step to validate that RPC methods follow a naming convention specified by the user. If we wanted to we could default it to the convention we're providing, but then we're not boxing users into a convention they may not want.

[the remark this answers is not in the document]
TJHeuvel: Cant users do that already with Roslyn analysers, or EditorConfig with custom rules?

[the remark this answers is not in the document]
becksebenius-unity: Yes they can, this alternative would just be to provide the validation functionality without having to build it yourself

PhantomCloak:
in my opinion Entity Framework use specs is not completely fit the definition of “Standard C#” maybe in the first glance, yes

but in complex operations sake of Entity Framework's use specs sacrifices too much on “Standard C#” definition, for example, you can work with anonymous types in query operations which are the exact opposite of strongly typed language definition, like the example below

`var items = context.Account.Select(row => new { IdMember = row.Email, UserName = row.Password }).ToArray();`

in this example, you cannot define `var items` explicitly because it’s an anonymous type so even though it’s a compiler feature 

The “Standard C#” term breaks down here if I’m not mistaken, right? and there is much more example like this sql-ish linq, pattern matching, anonymous types goes on and on it would apply same when Roslyn code generator feature comes in thus I think the “Standard C#” definition is too flexible than we think.

MFatihMAR:
@TwoTenPvP @LukeStampfli 

First of all, I will inline this one more time again:

> I'm happy to consider this discussion in a separate RPC API RFC as a successor but so far, it all sounds like personal preferences to me.

Let me put another spotlight on this:

> > but I can tell you that direct function call without extra framework utilities in the middle is the most loved and embraced option in the industry

> Where is the source for that? People like the MLAPI for the exact opposite reason.

I'm not sure if you read my comment entirely, I'm pointing you towards Unreal RPCs, I also called out that in the RFC document itself but still, let me paste you a link for that as well: https://docs.unrealengine.com/en-US/InteractiveExperiences/Networking/Actors/RPCs/index.html

Beyond that, the source is AAA engines and games, maybe I can give you another link: https://youtu.be/L4m2nwn5wT8?t=1431
I can't give you too much links, you had to be there first to see what's going on but I'm coming from that background and I can tell you that, people even do post-compile codegen/macrogen to have these goodies, we're achieving the same result even more conveniently.

It's not "regular C#" argument is weak, really. Unity already does lots of things such as custom equality operator, DOTS Entities etc. Microsoft also does that, just a few minutes ago somebody pointed out anonymous types being a "regular C#" because Microsoft just wanted to have that new "syntactic sugar" which helps people. I can follow up on this topic if you want, with more and more cases and concrete examples where "regular C#" is being violated from your perspective by having some framework features doing something behind the scenes. I can give you some spoilers: Roslyn Source Generators, ASP.Net Dependency Injection, EF Core feaatures and lots of other Microsoft products "breaking regular C#" and "how C# works". If there is no code you wrote to create instances of types you specified in the constructor but they somehow instantiated and passed onto your constructor, how that's "regular C#"? Find me the C# spec that specifically covers a DI framework design and its implementation. Surprise, both Unity and Microsoft broke C# in that regard already.

Let me inline what ILPP produces before going for my other bigger follow-up comment:

```cs
public class ExpStdRPC : NetworkedBehaviour
{
    [ClientRPC]
    private void BarClientRPC(int framekey, ClientRPCOptions options = default)
    {
        var singleton = NetworkingManager.Singleton;
        if (singleton == null || !singleton.IsListening)
        {
            return;
        }
        if (__nexec != NExec.Client && (singleton.IsServer || singleton.IsHost))
        {
            var bitWriter = BeginClientRPC(options.TargetClientIds);
            if (bitWriter != null)
            {
                bitWriter.WriteUInt32Packed(123123);
                bitWriter.WriteInt32Packed(framekey);
            }
            EndClientRPC(bitWriter, options.TargetClientIds);
        }
        if (__nexec == NExec.Client && (singleton.IsClient || singleton.IsHost))
        {
            print($"{nameof(BarClientRPC)} -> {framekey} --- {options.TargetClientIds.Length}");
        }
    }
}
```

If you think the code above is not "regular C#" because ILPP, Roslyn SG or any other source code generator generates it for you, that'd be really strange to me.



@LukeStampfli 

Disabling a part of a framework (ILPP in this case) and expecting rest of it working fine is strange. Would you expect a crucial part of ASP.Net being disabled by continue working just fine because it has to?

Also, why it's OK to ILPP everything except the proposed API?



C# does not come with network construct like RPC, we're providing that by Unity/C#, same as very basic DI framework or some other syntactic sugars (async-await, tuples etc etc) Microsoft thought might be handy to work with. Why it's OK for Microsoft to have frameworks, products and syntactic sugars but it's not OK for us to provide those?

C# is not the only language out there with no built-in network constructs but there are also some other languages with network constructs such as Erlang (and Elixir built on top of Erlang): https://erlang.org/doc/man/rpc.html
The language itself provides RPC mechanism by the language spec itself, you don't have to bring that in. But on C# side, it's not a language with network constructs by default, we bring them into C# and that's OK.

There could be arguments made over which one might be more explicit, clear etc etc but even in that regard, we could say we want functions to be called like `InvokeHelloWorldServerRpc()` rather than `HelloWorldServerRpc()` and that'd be also fine. The whole "we're breaking regular C# here" argument just doesn't add-up well.



Once again:

> I'm happy to consider this discussion in a separate RPC API RFC as a successor but so far, it all sounds like personal preferences to me.

text/0000-std-rpc-api.md
MFatihMAR: @JashanChittesh

Deterministic hashing of RPC names means just hash the method name, that's not a great idea, your method signature could change therefore you actually need to change your serialization code and not resolve into same RPC serialization code. Having said that, we will be backward-compatible as much as possible, [I inlined some more details here](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/std-rpc-api/text/0000-std-rpc-api.md#backward-compatibility).

It is fundamentally:

> Framework registers RPC methods statically by their deterministic [method signature](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/methods#method-signatures) hash which means as long as method signature stays the same, backward-compatibility will be possible but if method signature changes, same method with the same name under same namespace and type will not be compatible with its predecessor.

I will also inline other cross-compatibility sections very soon, thanks for raising this up!

text/0000-std-rpc-api.md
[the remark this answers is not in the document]
JoNax97: I would also prefer to leave this as a best practice, backed by IDE hints, Roslyn and the like.

paulpach:
This looks very similar to MirrorNG,  we have [ServerRpc] and [ClientRpc] and they are allocation free.  

The way we do it is:  when we find a [ServerRpc] method,  we replace it with a stub and we generate a corresponding skeleton via IL post-processing with Cecil.  We move the user's code into a new function.  This way we don't have to replace the calls to the Rpc everywhere.  We do not impose any naming convention on the method.

If it helps, here is the code that generates the stub:
https://github.com/MirrorNG/MirrorNG/blob/436f53c92a5469c41a2b1292a5eb70a32a4a5f44/Assets/Mirror/Editor/Weaver/Processors/ServerRpcProcessor.cs#L54

and here is the code that generates the skeleton:
https://github.com/MirrorNG/MirrorNG/blob/436f53c92a5469c41a2b1292a5eb70a32a4a5f44/Assets/Mirror/Editor/Weaver/Processors/ServerRpcProcessor.cs#L142

NoelStephensUnity (approved):
MD looks great, the discussion was an interesting read (liked your final summary on the C# rules vs codegen etc).

text/0000-std-rpc-api.md
```diff
@@ -0,0 +1,503 @@
...
+void MyUnreliableClientRpc() { /* ... */ }
+```
+
+Reliable RPCs will be received on the remote end in the same order as they are fired but this in-order guarantee only applies to RPCs on the same `NetworkObject`. Different `NetworkObject`s might have reliable RPCs called but executed in different order compared to each other. To put more simply, in-order reliable RPC execution is guaranteed per `NetworkObject` basis only.
+
+An RPC call made without an active connection will be dropped and will not be queued for send automatically. Both reliable and unreliable RPC calls have to be made when there is an active network connection established between a client and the server. Also reliable RPC calls made during connection will be dropped on disconnect as well.
+
+### Execution Table
+
+An RPC function **never** executes its body immediately since it's being a network construct. Even a `ServerRpc` called by a host (an instance that is a client and the server at the same time, aka listen server) will not be executed immediately but follow the regular network frame staging first.
```
mattwalsh-unity, line 160: suggest: "An RPC function never executes its body immediately since the function call really is a stand-in for a network transmission"

text/0000-std-rpc-api.md
```diff
@@ -0,0 +1,503 @@
...
+
+||Server|Client|Host (Server+Client)|
+|-:|:-:|:-:|:-:|
+|ServerRpc Network Send|❌|✅|✅|
```
mattwalsh-unity, line 164: I had to really think about this table.  I think this could be illustrated more clearly maybe more of a ladder / flow diagram where you have the server / client as the 'columns' and have time going vertically

text/0000-std-rpc-api.md
```diff
@@ -0,0 +1,503 @@
...
+|ClientRpc Network Call|❌|✅|✅|
+|ClientRpc Direct Call|❌|❌|❌|
+
+## RPC Params
+
+Both `ServerRpc` and `ClientRpc` methods can be configured either by `[ServerRpc]`/`[ClientRpc]` attributes at compile-time and/or `ServerRpcParams`/`ClientRpcParams` at runtime.
+
+Developers can put `ServerRpcParams`/`ClientRpcParams` as the last parameter (optionally) and they could be used for a consolidated space for `XXXRpcReceiveParams` and `XXXRpcSendParams`.
+
+The network framework will inject `XXXRpcReceiveParams` when invoked by network receive handling (framework code) and will consume `XXXRpcSendParams` when invoked by RPC send call (user code).
```
mattwalsh-unity, line 177: Suggest: The network framework will inject the corresponding XXXRpcReceiveParams to what the user had declared in code when invoked by network receive handling (framework code) and will consume XXXRpcSendParams when invoked by RPC send call (user code).

text/0000-std-rpc-api.md
```diff
@@ -0,0 +1,503 @@
...
+
+In short, developers will see error and warning messages in the Unity Editor in the Console window if there was a problem with RPC method definition, parameters types on RPC method signature etc.
+
+Some of those messages will be warning messages which will not prevent compile and/or build.
+
+Some of those messages will be error messages which will prevent compile and/or build.
+
+Descriptive messages, message types, when messages should appear and other details are very much dependent on the implementation of this RPC API specification, therefore they will be decided while implementing the feature.
+
+### Roslyn Unity Analyzers
```
mattwalsh-unity, line 347: Can we mention "Roslyn feature rollout will occur in sync with Roslyn availability in release Unity versions"

mattwalsh-unity (approved):

text/0000-std-rpc-api.md
MFatihMAR: Actually, Roslyn Analyzers are supported in Unity today (but Roslyn Source Generators is not). Also, in the "Strategy" section below, we're only highlighting Unity Diagnostics implementation and leaving both Rider Unity Plugin and Roslyn Unity Analyzers for the future and never promise a specific timeline or commitment.

text/0000-std-rpc-api.md
```diff
@@ -0,0 +1,503 @@
...
+```cs
+// Error: Invalid, missing 'ServerRpc' suffix in the method name
+[ServerRpc]
+void Ping(int somenumber, string sometext) { /* ... */ }
+
+// Error: Invalid, missing [ServerRpc] attribute on the method
+void PingServerRpc(int somenumber, string sometext) { /* ... */ }
+```
+
+`[ServerRpc]` attribute and matching `...ServerRpc` suffix in the method name are there to make it crystal clear for RPC call sites to know when they are executing an RPC, it will be replicated and executed on the server-side, without necessarily jumping into original RPC method declaration to find out if it was an RPC, if so whether it is a ServerRpc or ClientRpc:
```
CDMayberry, line 80: On requiring ServerRpc/ClientRpc suffixes: I'd prefer this not be enforced. MS doesn't enforce that all events end in "Event" nor delegates in "Delegate", and they only encourage that Async functions end in "Async" but also don't require it. If I have a function I want to write to be a ClientRpc, but on top of that I'm going to have only the owner use it, I'd rather call it PongAuthority or PongOwner than PongClientRpc. Vice versa, if I want everyone who wasn't a authority to receive a ClientRpc, I'd call it PongObservers or the like.

As Paul noted, the MirrorNG crew has removed this requirement and I haven't seen it become a problem. Beginner's in C# might make the mistake of improperly naming it, but they're probably having that issue with more than just RPC method names. IMO it'd be better to point them to a guide about good naming conventions in the docs than force everyone to use a specific naming convention.

MFatihMAR: Hey @CDMayberry, thank you for the feedback!

I think we just simply want to play a little safe here and maybe that would also mean we wouldn't require API-breaking changes in the future, let me explain a few things.

UNET already used some prefixes in the past (see [UNET Remote Actions: Command and ClientRpc](https://docs.unity3d.com/Manual/UNetActions.html)) — AFAIK, it never got a huge backlash and people were generally OK with it.

We do care about beginners and we want to cover them in all cases if we can (also proper naming is subtle and even experienced developers are falling into traps occasionally). However, I agree that it's a little bit too strong enforcement and we might reconsider this decision again (see "Unresolved questions" and "Future possibilities" sections in the RFC markdown document).

Playing restrictive also gives us an advantage on API future-proofing: we might weaken enforcement rules but that wouldn't require any API changes but strengthening enforcement rules would.

I think we'll be more than happy to reconsider this again, especially after implementing extensive tooling (Roslyn Analyzers, Rider Unity Plugin) and/or hearing from the community around this enforcement causing undesired issues etc.

Before I wrap-up, let me inline a few parts from the RFC document here:

from "Unresolved questions" section:

>  - Weakening `...ServerRpc`/`...ClientRpc` suffix enforcement
>    - Should they really be hard compile errors or just soft warnings?
>    - Should we allow suffix syntax configuration?
>    - Should we allow suffix, prefix or none configuration?

from "Future possibilities" section:

> We might follow-up on Unresolved questions to address with further implementations

I hope this makes sense :)

text/0000-std-rpc-api.md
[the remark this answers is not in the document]
mattwalsh-unity: I won't veto the process over it, but for the record I would prefer to not force the suffix.  I guess it's the language designer part of me that feels that devs cement their mental models & understanding when they see the 'why' of each piece of syntax.  I think folks will be consufed on why you need both the attribute and the suffix - it will lead them to wonder if they are doing 2 disparate things each handled by one of the 2.

mattwalsh-unity (approved):

text/0000-std-rpc-api.md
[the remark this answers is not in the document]
MFatihMAR: I will create another RFC & PR that suggest relaxing suffix enforcement but also PR will be implementing Roslyn Analyzers so that it would be still a pleasant experience for developers to see warnings/hints suggesting a better naming convention. (My main argument was to prevent people from making silly mistakes and there is no Roslyn Analyzers implementation yet but that'd change very soon with Unity2020.2 🎉)

text/0000-std-rpc-api.md
CDMayberry: > UNET already used some prefixes in the past (see UNET Remote Actions: Command and ClientRpc) — AFAIK, it never got a huge backlash and people were generally OK with it.

My comment is the backlash from using it for so long XD. I've worked with UNet/then Mirror for 5 years, my comment about PongAuthority vs PongObservers was from experience.

> from "Unresolved questions" section:

Ah my bad, I'll shift this to future discussions if/when the RFC pops up.

JoNax97:
I general I like the proposal. I'm not a fan of the naming convention but I trust it to get better in the future with Roslyn and good IDE support. 

On the other hand I can't agree with the loss of return values. It feels like all the work made to make RPCs look and feel more like regular methods is thown off by the heavy boilerplate of the RPCparams.

My ideal version of RPCs with return type would be to await them like any async method, and use them like this:

var value = await  GetSomethingFromServerRPC();

PR #2, opened 2020-12-06
Serializable Types

MFatihMAR:
[See rendered RFC document &rarr;](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/type-serialization/text/0002-serializable-types.md)


text/0000-serializable-types.md
```diff
@@ -0,0 +1,122 @@
...
+
+Static arrays like `int[]` will be serialized by built-in serialization code if their underlying type is either one of serialization supported types (e.g. `Vector3`) or if they implement `INetworkSerializable` interface.
+
+### Generic Collections
+
+Generic collections such as `IEnumerable<T>` and `IEnumerable<KeyValuePair<K, V>>` will be serialized by built-in serialization code if their underlying type is either one of serialization supported types (e.g. `Vector3`) or if they implement `INetworkSerializable` interface.
+
+### INetworkSerializable
+
+Complex user-defined types that implements `INetworkSerializable` interface will be serialized by user provided serialization code:
```
TJHeuvel, line 45: I dont know if this document is ready for discussion.

But i think there are two ways of doing it, having to implement an interface or registering serialization at some point. Currently [MLAPI does both](https://mlapi.network/wiki/custom-serialization/), which is confusing.

I think one important difference between the two is that when registering a custom serializer-per-type you have a lot more control. You can serialize types in assemblies you dont own, e.g. i buy a DLL from the asset store and i cannot access the code to make it implement INetworkSerializable. Or Unity types, i can register a serializer for a RigidBody and sync these with my custom method. When registering it is possible to add this functionality to existing classes, with interfaces it is not.

Another case is overwriting default behaviour. I might want to override how Unity Multiplayer sends `float`, because i know that in my game the world is much smaller, so i can get away with sending a `short` instead.

MFatihMAR: you could always implement `MyExternalTypeSerializable : MyExternalType, INetworkSerializable` or simply put `MyExternalType` under another `INetworkSerializable` type and serialize as a part of it.

_(this document is not ready, just created to extract serialization topic out of RPC)_

text/0002-serializable-types.md
```diff
@@ -0,0 +1,413 @@
...
+void MyServerRpc(NetworkObject nobj, NetworkBehaviour nbhv) { /* ... */ }
+
+void Update()
+{
+    if (Input.GetKeyDown(KeyCode.P))
+    {
+        MyServerRpc(/* NetworkObject = */ this.NetworkObject, /* NetworkBehaviour = */ this); // Client -> Server
+    }
+}
+```
```
MFatihMAR, line 119: (sigh) this section still seems cryptic to me...

text/0002-serializable-types.md
```diff
@@ -0,0 +1,413 @@
...
+[ServerRpc]
+void HelloServerRpc(int[] scores, Color[] colors) { /* ... */ }
+
+[ClientRpc]
+void WorldClientRpc(MyComplexType[] values) { /* ... */ }
+```
+
+### NetworkObject & NetworkBehaviour
+
+`NetworkObject` and `NetworkBehaviour` instances will be serialized by built-in serialization code if instances are not `null` and `NetworkObject.IsSpawned == true`. Passing a `null` instance or a `NetworkObject` with `IsSpawned == false` or a `NetworkBehaviour` with `NetworkObject.IsSpawned == false` will still lead into RPC replication but remote side will have `null` values for the parameters. IDs of spawned `NetworkObject` and `NetworkBehaviour` instances will be resolved by running `NetworkManager` instance, and those IDs will be the links between local and remote instances. Also, those IDs will be used when serializing `NetworkObject` and `NetworkBehaviour` instances as a part of an RPC call.
```
mattwalsh-unity, line 106: So I'm assuming it's not so much that the NetworkObject / NetworkBehaviour types are serialized (e.g. their contents are packed / unpacked in the RPC like the previous types).  Rather, you are transmitting to the recipient that this parameter is a NO / NB, and here's its ID that gets looked up by the NetworkManager and de-referenced on the receiving end.   That is, the actual transmission of the NO / NB state is happening through the NetworkingManager's Netvar mechanism, and this is just a way to pass a reference to a NO / NB.

If I have this right, I might suggest putting the mechanism like I described it first as when I first read this I thought "how is the serialization system magically serializing arbitrary types in my NO / NB without an INetworkSerializable interface.

MFatihMAR: Yes, it's just ID-linking on the remote side and dereferences back to an already spawned NetworkObject and/or NetworkBehaviour. I will rework this section with better wording and outline the actual mechanism as you suggested!

text/0002-serializable-types.md
```diff
@@ -0,0 +1,413 @@
...
+    if (Input.GetKeyDown(KeyCode.P))
+    {
+        MyServerRpc(/* NetworkObject = */ this.NetworkObject, /* NetworkBehaviour = */ this); // Client -> Server
+    }
+}
+```
+
+### INetworkSerializable & BitSerializer
+
+Complex user-defined types that implements `INetworkSerializable` interface will be serialized by user provided serialization code.
```
mattwalsh-unity, line 123: Nitpick, implements -> implement

text/0002-serializable-types.md
```diff
@@ -0,0 +1,413 @@
...
+
+An instance of `BitSerializer` will be passed into `INetworkSerializable::NetworkSerialize(BitSerializer)` method which can be used to easily serialize fields by reference.
+
+All types supporting serialization will also be supported by `BitSerializer` with `BitSerializer::Serialize(ref int value)` variant methods and templated `BitSerializer::Serialize<T>(ref T value) where T : INetworkSerializable` method.
```
mattwalsh-unity, line 127: You mention the 'int' variant method but not the 'float' variant.

MFatihMAR: oh, I mean all variants of built-in supported types. I will clarify.

text/0002-serializable-types.md
```diff
@@ -0,0 +1,413 @@
...
+
+```cs
+class BitSerializer
+{
+    bool IsReading { get; }
+
+    void Serialize(ref int value) { /* ... */ }
+    void Serialize(ref float value) { /* ... */ }
+    // ...
```
mattwalsh-unity, line 136: I assume here by ... you mean the other built-in types (e.g. Vector3).  If so, I'd say so.  I had to stop and think for a moment why it had just int and float here

MFatihMAR: yep, all primitives listed above including string, Vector3, enum etc. I think I should put more hints around shortenings to clarify intentions.

text/0002-serializable-types.md
```diff
@@ -0,0 +1,413 @@
...
+    public void NetworkSerialize(BitSerializer serializer)
+    {
+        // Length
+        int length = 0;
+        if (!serializer.IsReading)
+        {
+            length = Array.Length;
+        }
+
+        serializer.Serialize(ref length);
```
mattwalsh-unity, line 198: Pretty obvious, I know, but probably worth saying "Encode the length"

text/0002-serializable-types.md
```diff
@@ -0,0 +1,413 @@
...
+
+```cs
+[ServerRpc]
+void HelloServerRpc(int[] scores, Color[] colors) { /* ... */ }
+
+[ClientRpc]
+void WorldClientRpc(MyComplexType[] values) { /* ... */ }
+```
+
+### NetworkObject & NetworkBehaviour
```
becksebenius-unity, line 104: Are there ever circumstances where a network id could be passed via RPC or custom serialization before the network object has actually been spawned? This is less important for RPCs I think and more about how this serialization strategy applies to network vars / state. Because for example if one network behavior references another, depending on the order of spawning them the reference may not be able to be fulfilled because the object does not yet exist. Or, does the network spawning always happen before state deserialization? I'm not familiar with the internals but just curious if this type of ordering has been considered when it comes to object references

text/0002-serializable-types.md
becksebenius-unity, line 136: How does enum work with this strategy? in my experience, serializing user enums almost always leads to boxing because c# still (much to my lament) does not support enums as primitive generics. (as in, you can't say `where TEnum : int`. In my experiments with this pattern, I've had to do something like this:
```int myEnumValue = (int)myEnum;
serializer.Serialize(ref myEnumValue);
myEnum = (MyEnum)myEnumValue```

which is pretty verbose.

File: text/0002-serializable-types.md [hunk 9]
```diff
@@ -0,0 +1,413 @@
...
+    // for reading
+    BitSerializer(BitReader reader) { /* ... */ }
+
+    // for writing
+    BitSerializer(BitWriter writer) { /* ... */ }
+
+    bool IsReading { get; }
+
+    // for common types
+    void Serialize(ref int value)
```
becksebenius-unity, line 350: Have you thought at all about how bit packing might work with this interface? Is that just up to the user before sending to the serializer?

text/0002-serializable-types.md
```diff
@@ -0,0 +1,413 @@
...
+}
+```
+
+#### Conditional Serialization
+
+As the developer has more control over serialization of a struct, one might implement conditional serialization at runtime.
+
+We will explore more advanced use-cases with the examples below:
+
+##### Example: Array
```
mattwalsh-unity, line 182: I guess I wonder, could we have a built-in array?  if the array is just a sequence of INetworkSerializable types, then could we give users out-of-the-box array support vs. having to write an array handler?

becksebenius-unity: +1, this seems like the type of code that could be generalized in BitSerializer. Although it adds another layer of serialization code needed for each type supported which can't really be encapsulated

text/0002-serializable-types.md
```diff
@@ -0,0 +1,413 @@
...
+An instance of `BitSerializer` will be passed into `INetworkSerializable::NetworkSerialize(BitSerializer)` method which can be used to easily serialize fields by reference.
+
+All types supporting serialization will also be supported by `BitSerializer` with `BitSerializer::Serialize(ref int value)` variant methods and templated `BitSerializer::Serialize<T>(ref T value) where T : INetworkSerializable` method.
+
+```cs
+class BitSerializer
+{
+    bool IsReading { get; }
+
+    void Serialize(ref int value) { /* ... */ }
```
mattwalsh-unity, line 134: True / False: in the end, all reads/writes to BitSerializer will be done through one of the built-in types.  From the examples and my understanding, even if I have a custom type, my serialization operations will just be a sequence of built-in type operations.

If I have this right, it would mean you could not, say, do weird bitwise packing (like pack a bool into a byte or even a bit).  I'm not saying that's a bad thing, IIRC @pandr & his group found this to be not worth the bandwidth savings.  But I would have liked this called out earlier on to help me get that insight sooner.

If I have this wrong, I'd suggest showing an example (presumably like how you'd get to m_writer in BitWriter).

text/0002-serializable-types.md
MFatihMAR: I think MLAPI doesn't guarantee such consistent today but by sheer luck, it works fine most of the time (don't quote me for that). There is a bug SF guys raised internally a few days ago which I believe is related to spawning/referencing/replication issue you're talking about here. For now, it's unclear with no guarantees. Unless @TwoTenPvP knows better than me in this area.

text/0002-serializable-types.md
MFatihMAR: `BitSerializer` is a flip-flop container that holds `BitReader` and `BitWriter` internally. Bit-packing and other serialization optimization techniques should be implemented over there (and there are some packed stuff already).

text/0002-serializable-types.md
MFatihMAR: we will provide static array serialization out-of-the-box as outlines above in this RFC, this array example is just serving a demonstration purpose here.

text/0002-serializable-types.md
mattwalsh-unity: Today, at least for Network Vars, MLAPI checks if a networkId is valid to the recipient and if not, spawns a warning and, under circumstances involving message buffering, saves it up and tries again later (e.g. HandleNetworkedVarUpdate)

Similarly in BitReader.cs, when it's dealing with a game object (in ReadObjectPacked) it throws out a warning.

I don't think you'll every catch all the cases regardless of spawn order / deserialization because clients can wink out at any time.  So at first blush I would be skeptical about trying to make this water-tight / make it the serialization system's responsibility.

HOWEVER I would be very interested in exploring having a clean mechanism for user code to get a definable callback for when you de-serialize and find the NO doesn't exist on your side vs. the warning today.

text/0002-serializable-types.md
MFatihMAR: yes, you got it right but there are some other things worth mentioning here. `BitSerializer` will not expose `BitWriter` and `BitReader` it holds internally but it will expose `Serialize(ref byte value)` and `Serialize(ref byte[] value)` API which could be used for any kind of manual serialization including bit packing, quantization and other compression techniques. I think I should inline the whole `BitSerializer` API here to clarify what's being proposed here (looks like some things stuck in my mind but not written entirely here)

text/0002-serializable-types.md
becksebenius-unity: Yeah, I would be curious to see how circular references work for join-in-progress scenarios since one side of the reference would always be invalid until all objects are finished spawning. The buffering does sound like a pretty elegant solution for that case though. That said, this concern might be out of scope for this RFC

text/0002-serializable-types.md
MFatihMAR: C# 7.3 FTW! `void Serialize<TEnum>(ref TEnum value) where TEnum : Enum`

becksebenius-unity: Hmmm does that work though? I haven't actually used that clause but my understanding reading the release notes (to be fair, before it actually released) was that it didn't actually have teeth - you couldn't generically convert to the backing data type, which means you have to use APIs like System.Convert which will box the enum, rather than doing things like `(int)myEnum` (compiler error because the generic method can't guarantee that it's an `int` enum)

text/0002-serializable-types.md
mattwalsh-unity: Great, yeah I think just more or less pasting in what you just described would be great just to indicate you can do such trickery

text/0002-serializable-types.md
MFatihMAR: you can get underlying enum type using `Enum.GetUnderlyingType(Type)` API — in fact, that's also in the RFC document, see `BitReader` section down under the "Reference-level explanation" title.

text/0002-serializable-types.md
```diff
@@ -0,0 +1,553 @@
...
+            /* NetworkObject = */ this.NetworkObject,
+            /* NetworkBehaviour = */ this);
+    }
+}
+```
+
+Serializing `NetworkObject` and/or `NetworkBehaviour` doesn't mean their fields will be serialized. Only IDs will be serialized and remote-side will lookup by IDs locally to retrieve instances. Serialization of these types fundamentally links local and remote instances via IDs.
+
+```cs
+class BitSerializer
```
LukeStampfli, line 126: Adding fholms comment here for visibility.

BitSerializer should really just be a struct instead of a class because it just wraps BitReader/Writer.

MFatihMAR: yep, agreed! we can make it a `struct` on actual implementation side. RFC won't enforce it to be a `class` :)

dwoodruffsf (commented):
I have a couple of additional comments:

* At Turbine, we also used a model with a single `Serialize` method for both reading and writing. We can endorse it, but thought it was worth mentioning that it does have an intrinsic performance cost, because the serialization code is more branch-y (essentially an `if else` wrapping every leaf in the serialization tree). 

* Another potential performance issue is whatever dynamic array solution the Stream is wrapping. Constantly growing dynamic arrays can be expensive in allocations. Other networking solutions let you pre-size message buffers to prevent this. This is probably a separate issue from serializable types, but might be something to consider for the RPC implementation (i.e. some way to hint in code that an RPC needs a 20 byte buffer vs a 1MB buffer). 

* Finally, I wanted to call out a problem that will show up as soon as we start working on the MMO demo. We're going to have some `CharacterState` object, and it's going to need to be serialized on the wire (from server to client on login). It's _also_ going to be persisted to a database. It would suck to have that object need two separate serialization flows. That means it will be important to extend this serialization model to support versioning. 

The most important implication for your implementation is that once you've added serialization for a type, you won't be able to change it, except by adding a new "Serialize[Type]V2" method. You can mainly leave it to users to handle the nuts and bolts of versioning (serialize a version integer at the start of the stream). However, a nice user-feature would be to be able to push some context onto the BitSerializer object, so if you have one parent object with 10 sub-objects, you don't necessarily need to serialize 10 sub-versions (because they can be inferred from the parent's version).

text/0002-serializable-types.md
```diff
@@ -0,0 +1,553 @@
...
+[ServerRpc]
+void HelloServerRpc(int[] scores, Color[] colors) { /* ... */ }
+
+[ClientRpc]
+void WorldClientRpc(MyComplexType[] values) { /* ... */ }
+```
+
+### NetworkObject & NetworkBehaviour
+
+`NetworkObject` and `NetworkBehaviour` instances will be serialized by built-in serialization using their respective network object ID (`NetworkObject.NetworkId`) and network behaviour ID (`NetworkBehaviour.GetBehaviourId()`).
```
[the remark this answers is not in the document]
MFatihMAR: very good point and I think I agree we should either serialize them or just not offer them and leave it up to dev to send `NetworkId` over if needed. what this proposes it sub-par. this feature is not new, it's actually migrating from existing MLAPI and making improvements on top of that (UNET also offered similar functionality). I think it's fair to say that we should leave this for later rather than putting it in early without thinking too much about what it means and what it would communicate.

long story short: I agree that we should leave this out for now.

text/0002-serializable-types.md
```diff
@@ -0,0 +1,553 @@
...
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## Serialization
+
+Multiplayer framework has built-in serialization support for C# and Unity primitive types out-of-the-box, also with ability to further extend network serialization for user-defined types implementing `INetworkSerializable` interface.
+
+### C# Primitives
+
+`bool`, `char`, `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `float`, `double`, `string` types will be serialized by built-in serialization code.
```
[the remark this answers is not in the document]
MFatihMAR: this one could be something we'd address later as it's called out in the "Future possibilities" section at the bottom:
> **Templated Custom Serializers**
> Currently, there is no way to swap `BitSerializer` with something else. It is still possible to manually serialize types down to `byte[]` and use `BitSerializer.Serialize(ref byte[] value)` API but it would be much better to have full control over serializer. This RFC considers this as a potential improvement for the future.

on your top comment, I'd like to touch on a few things.

first of all, BitStream and other managed types around serialization will be replaced with NativeArray and unmanaged memory operations which are also going to be helpful when we move more and more towards jobified internals. for now, we're just sticking with BitStream/BitReader/BitWriter trio and adding BitSerializer to help us just a little more. performance-wise, what we have today is not ideal and a bit far from what we're planning for the future.

for the other major areas: custom serialization, extending with new types, having more manual control and all other will potentially be introduced with a new API but not necessarily under existing BitSerializer/BitStream/BitReader/BitWriter stuff. `Serialize[Type]V2` API won't ever become a reality. what will be happening though is that we will introduce new set of classes backed by NativeArray/NativeSlice which are going to be job-friendly, more modular & customizable and potentially offer much much more value add compared to what we have today and what we will get with this RFC.

in terms of planning and rollout, we will most likely to have BitXXX stuff and new cool stuff coming up later side by side for a while, up until we fully migrate into cool stuff everywhere around the whole framework (including internals + public facing APIs) with 1:1 matching functionality at bare minimum and potentially more stuff on top.


again, long story short here: I 100% agree on all the points you raised above and I think we will eventually get there. only thing to call out here is that this RFC makes existing stuff a little bit better but won't be the final solution. we want these changes to improve things today and we will make further changes in this area again in the future.

mattwalsh-unity (approved):

text/0002-serializable-types.md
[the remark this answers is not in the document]
LukeStampfli: David pointed me to this 😃. I disagree on some of the points which have been made.
1. NetworkedObject isn't a class which can be serialized. It is just an identifier, it basically acts like an entity in an ECS.
2. This is a problem which can be solved. Solving this problem with magic numbers can work but I'd argue it is not good design for a framework. 

I totally agree though on the on the lifetime argument and that way Unity is solving this with it's Objects isn't great.

One way I think to get the best of the two worlds would be the following:

1. We have a NetworkedObjectId struct. That struct contains just a ulong field. When sending NetworkedObjects we send/receive that struct. As a user I can try to receive a NetworkedObject from this Id struct.  Benefits of this approach:
- When we designate a special number like 0 to "not found" we can expose this in a nice way with a "Found" property.
- The check whether a NetworkedObject exists will be done by the framework and not by the user. This means less boiler plate code and more opportunities to log issues. BitWriter/BitReader currently contains a bunch of very useful logging statements about missing NetworkedObjects.
- It's more extensible/maintainable, if we ever decide to change NetworkId from ulong to int we can do that without breaking user code, they won't even notice.

I'm still not entirely sold on this mainly because UNet/Mirror/PUN all use this pattern of serializing object because it is just really convenient and intuitive.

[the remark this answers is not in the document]
MFatihMAR: @LukeStampfli — I actually do not have any strong opinions about this. However, I think it's out-of-scope for this RFC to standardize or propose a solution, that's why I'd like to leave it outside for now. (In fact, I implemented ID-based `NetworkObject` and `NetworkBehaviour` serialization over at MLAPI repo under `experimental/serializable-types` branch but removed it again). Let's have a separate discussion and RFC for this :)

PR #5, opened 2020-12-16
RPC Queue

NoelStephensUnity:
[See linked RFC document &rarr;](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rpcqueue/text/0005-rpc-queue.md)

First pass at getting an RFC in place for the RPC Queue stuff.


MFatihMAR:
I will trying to figure out how we could put our diagrams into markdown documents, so styling/formatting TBD :)

text/0000-rpc-queue.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Inbound RPC Data Flow Pipeline
+![](0000-rpc-queue/InboundDataFlowPipeline.png)
+
+### RPC Queue Classes
+![](0000-rpc-queue/RPCQueueClasses.png)
+
+- **RPCQueueManager:** Manages inbound and outbound RPC queues.  Both inbound and outbound queues are byte arrays exposed as streams (i.e. currently BitStreams) and contained within a QueueHistoryFrame.  The number of QueueHistoryFrames is defined by calling the Initialize method and passing the maximum number of history frames.  The total number of frames is MaxFrameHistory + 1, where the additional frame is considered the “current frame” (i.e. it will always maintain the exact number of frames in history).
+- **RPCQueueProcessing:** Currently, this class is instantiated by the RPCQueueManager during its initialization.  It is highly probable that this class will get absorbed into the RPCQueueManager.
+- **QueueHistoryFrame:** Container class for handling the management of an RPC queue.  One QueueHistoryFrame instance per data flow pipeline (i.e. inbound and outbound yields two QueueHistoryFrames).
```
MFatihMAR, line 43: I might be a little bit nit-picky here but I think we should call `RPCQueueManager` as `RpcQueueContainer` instead and potentially call `RPCQueueProcessing` as `RpcQueueProcessor`. "Manager" implies singletons and we probably would not want to implement this queue as a singleton but just a container implementation instead.

text/0000-rpc-queue.md
```diff
@@ -0,0 +1,87 @@
...
+5. Providing a queue for inbound messages provides an alternate level of control over when the message will be processed and in the case of RPCs provides the ability to synchronize RPC oriented messages that rely upon any tickrate based synchronization methodologies/implementations.
```
MFatihMAR, line 20: I feel like we're listing way too many things here and AFAIK the real motivation behind RPC queuing today is the fact that MLAPI calls `Send` on the transport immediately rather than giving us a control over how&when RPCs are being sent and received (currently all RPC receives are happening inside `NetworkingManager::Update` and RPC sends are being immediate Sends on the transport). Other suggestions and assumptions may or may not be there and I think they should not be directly related to this RFC's main motivation.

(I will also touch on a few other things below in the "Drawbacks" section)

text/0000-rpc-queue.md
```diff
@@ -0,0 +1,87 @@
...
+![](0000-rpc-queue/RPCQueueClasses.png)
+
+- **RPCQueueManager:** Manages inbound and outbound RPC queues.  Both inbound and outbound queues are byte arrays exposed as streams (i.e. currently BitStreams) and contained within a QueueHistoryFrame.  The number of QueueHistoryFrames is defined by calling the Initialize method and passing the maximum number of history frames.  The total number of frames is MaxFrameHistory + 1, where the additional frame is considered the “current frame” (i.e. it will always maintain the exact number of frames in history).
+- **RPCQueueProcessing:** Currently, this class is instantiated by the RPCQueueManager during its initialization.  It is highly probable that this class will get absorbed into the RPCQueueManager.
+- **QueueHistoryFrame:** Container class for handling the management of an RPC queue.  One QueueHistoryFrame instance per data flow pipeline (i.e. inbound and outbound yields two QueueHistoryFrames).
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we _not_ do this?
```
MFatihMAR, line 48: This section is meant for a very specific purpose as the template outlines: "Why we should _not_ do this?"

One thing immediately pops-up in my mind is framework-layer vs transport-layer trade-off.
Currently, transport layer is not very helpful when it comes down to determining a proper network frame (with frame start and frame end).

However, we might hold onto this RFC and revisit RPC queuing after we figure out a better transport-layer interface abstraction and implementation (with queuing highly encouraged -or forced- via transport interface API).
If that happens, the feature this RFC suggest might simply become obsolete and transport-layer queuing would effectively replace both RPC queuing and also all other potential queues as well.

So my ask here is to talk a little bit more about this trade-off and outline these decisions by picturing our current state in the project, also mention potential future work coming up on the transport layer so that we can come back to this RFC at any point in the future, understand the thought-process behind this feature RFC and/or move towards a better transport-layer abstraction with the given context from this RFC.

text/0000-rpc-queue.md
```diff
@@ -0,0 +1,87 @@
...
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+Think about what the natural extension and evolution of your proposal would be and how it would affect the Unity Multiplayer as a whole in a holistic way. Try to use this section as a tool to more fully consider all possible interactions with the Unity Multiplayer in your proposal. Also consider how the this all fits into the roadmap for the project and the team.
+
+This is also a good place to "dump ideas", if they are out of scope for the RFC you are writing but otherwise related.
+
+If you have tried and cannot think of any future possibilities, you may simply state that you cannot think of anything.
+
+Note that having something written down in the future-possibilities section is not a reason to accept the current or a future RFC; such notes should be in the section on motivation or rationale in this or subsequent RFCs. The section merely provides additional information.
```
MFatihMAR, line 87: following-up on the "Drawbacks" section, if we still decide to go for this after drawing a picture of the current state of the project, we might want to call out transport-layer rework here that could potentially make this RPC queuing feature proposal entirely obsolete or degrade its functionality at least at some degree.

MFatihMAR (changes requested):
Please see my comments above.

text/0000-rpc-queue.md
NoelStephensUnity: This was addressed in most recent updates to this particular RFC

text/0000-rpc-queue.md
```diff
@@ -0,0 +1,86 @@
...
+![](0000-rpc-queue/RPCQueueClasses.png)
+
+- **RPCQueueContainer:** Manages inbound and outbound RPC queues.  Both inbound and outbound queues are byte arrays exposed as streams (i.e. currently BitStreams) and contained within a QueueHistoryFrame.  The number of QueueHistoryFrames is defined by calling the Initialize method and passing the maximum number of history frames.  The total number of frames is MaxFrameHistory + 1, where the additional frame is considered the “current frame” (i.e. it will always maintain the exact number of frames in history).
+- **RPCQueueProcessing:** Currently, this class is instantiated by the RPCQueueContainer during its initialization.  It is highly probable that this class will get absorbed into the RPCQueueContainer.
+- **QueueHistoryFrame:** Container class for handling the management of an RPC queue.  One QueueHistoryFrame instance per data flow pipeline (i.e. inbound and outbound yields two QueueHistoryFrames).
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This approach does include additional processing overhead and there could be some refactoring depending upon any future transport related work that might occur.
```
MFatihMAR, line 64: do you mind elaborating a little more why there is an overhead and where it is coming from? also what we could potentially do about this in the future? if we have any rough plans/ideas, we might put them under "Future possibilities" section.

text/0000-rpc-queue.md
```diff
@@ -0,0 +1,86 @@
...
+- **QueueHistoryFrame:** Container class for handling the management of an RPC queue.  One QueueHistoryFrame instance per data flow pipeline (i.e. inbound and outbound yields two QueueHistoryFrames).
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This approach does include additional processing overhead and there could be some refactoring depending upon any future transport related work that might occur.  
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+Since there are no current discussions on batching specific packet types together at the transport layer (not recommended) and while Unet provides the ability to switch from immediate to queued mode there were other framework related design factors taken into condiseration (i.e. batched RPCs, invocation at different update stages, etc.) that made this the current best path to take.
```
MFatihMAR, line 68: why not comment (at least briefly) what we considered and why this approach suits our needs for now? also, an alternative might be to have `QueueSend()`/`FlushSend()`/etc. APIs on transport itself and expect it to happen under the hood for us, implemented by transport implementors. I'd like to see at least transport-level queuing vs framework-level queuing being mentioned here (especially since we have plans in the future in that direction already, why not note why this particular RFC is more useful than others today but potentially not tomorrow?).

text/0000-rpc-queue.md
```diff
@@ -0,0 +1,86 @@
...
+[unresolved-questions]: #unresolved-questions
+
+- What parts of the design do you expect to resolve through the RFC process before this gets merged?
+- What parts of the design do you expect to resolve through the implementation of this feature before stabilization?
+- What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+This provides the opportunity to place RPC's into batched packets (containing more than 1 RPC per packet) prior to being sent as well as it offers the ability to add additional features to specify at which point in the Network Update Loop system an RPC could get invoked.
```
MFatihMAR, line 86: Similar to points you briefly mentioned above (possible refactorings, changes in design etc.) could also be outlined here in separate sub-titles.

text/0000-rpc-queue.md
```diff
@@ -0,0 +1,86 @@
...
+[prior-art]: #prior-art
+
+
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- What parts of the design do you expect to resolve through the RFC process before this gets merged?
+- What parts of the design do you expect to resolve through the implementation of this feature before stabilization?
+- What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
```
MFatihMAR, line 81: we want to eliminate memory footprint and overhead we currently have but it's out of scope for this RFC, right?

NoelStephensUnity: Other than the pre-allocation of the BitStreams for each QueueHistoryFrame, after Matt's updates to the Channel Name (i.e. string to byte) RPCQueueContainer's footprint is close to zero (if not zero).  I still need to run further tests on this, but it seems the only things causing memory allocations is when entities are created.  One can run a deep profile on it and find frames where no entities are being created/destroyed and the GC Alloc is considered zero for RPCQueueContainer and associated classes.

text/0000-rpc-queue.md
NoelStephensUnity: Added a blurb on this in the most recent refactor.  I held back on going into any specific details as we still need some additional clarity regarding what we are going to expect from the default transport used with MLAPI (i.e. what the default Unity transport will end up being) and any changes to the framework transport interface that may be needed.

text/0000-rpc-queue.md
NoelStephensUnity: Added some additional possibilities to this area.  Still a bit vague and probably could stand to be further vetted.

text/0000-rpc-queue.md
NoelStephensUnity: The most recent version further addresses this area of the RFC.

text/0000-rpc-queue.md
NoelStephensUnity: The most recent version should address this area further.

MFatihMAR (approved):

PR #6, opened 2021-01-20
RPC Processing Hook

NoelStephensUnity:
There still needs to be some additions to the drawbacks and future possibilities, but for the most part this describes the NetworkUpdateManager registration process as well as how to specify when an RPC should be invoked on the receiving end.
Making this a PR at this point to get initial reactions as well as open discussion for any additions and/or changes.


MFatihMAR:
I still need to have another thorough look at this but I wonder if we should separate RPC Queue concerns out of this Network Loop System design RFC? Yes, an immediate use-case could be our current RPC Queue implementation but the main design this RFC proposes should still be just Network Loop stuff in isolation.

jeffreyrainy (approved):

PR #7, opened 2021-01-29
Performance Data Propagation

kvassall-unity:
[See linked RFC document →](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/performance-data/text/0007-performance-data-propagation.md)


text/0003-performance-data.md
```diff
@@ -0,0 +1,83 @@
...
+As this data is largely a struct like object, accessing its data will be as simple as accessing fields. Additionally the rendering or logging of these statistics will fall on the user and not affect the Tick or be done at all within the MLAPI framework.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+
+In order to limit the interface, we are pushing the collected set or performance data to be obtainable only at the end of a tick. The following is roughly what this object could look like. Currently this object will be updated in ```NetworkManager.LateUpdate()``` for the MLAPI and ```Transport.Send()``` and other relevant transport method for the transport layer. At the end of the tick the user will be notified that the relevant data is complete and ready (both the MLAPI data and Transport data).
+
+```cs
+class ProfilerTickData {
```
becksebenius-unity, line 35: We should make sure that this is done in such a way where no additional allocations are needed to gather this data, either by re-using the same object each frame or by using structs. This would inform the technical approach to gathering the data

text/0003-performance-data.md
```diff
@@ -0,0 +1,83 @@
...
+    int bytesReceived;
+    int tickDuration;
+    ...
+
+    // mlapi-level
+    int numberOfRPCs;
+    ...
+
+    // potentially? For any remaining edge cases?
+    Dictionary<string,Object> kvStore;
```
becksebenius-unity, line 49: This could be useful but I'd leave it as a future possibility rather than part of the initial implementation

text/0003-performance-data.md
```diff
@@ -0,0 +1,83 @@
...
+}
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+With a defined struct, we risk needing to know in advance as many fields and points of data that we want to push out of the Transport and MLAPI layer.
+Changes to the ProfilerTickData inherently cause changes to the transport, MLAPI, and all third party consumers.
```
becksebenius-unity, line 57: At least with this approach, adding new data types does not cause compatibility / compiler issues with 3rd party consumers

text/0003-performance-data.md
```diff
@@ -0,0 +1,83 @@
...
+
+    // potentially? For any remaining edge cases?
+    Dictionary<string,Object> kvStore;
+}
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+With a defined struct, we risk needing to know in advance as many fields and points of data that we want to push out of the Transport and MLAPI layer.
```
becksebenius-unity, line 56: I'd recommend updating the RFC to have intentional usage of struct and class. I think in this line you're really just referring to a data structure, not necessarily declaring that it is a struct or class. But if you _are_ making such a declaration then I think there should be a section about how this data is fulfilled given that structs are passed by value.

text/0003-performance-data.md
```diff
@@ -0,0 +1,83 @@
...
+
+External Tools and APIs will be able to then get this fully written ProfilerTickData at the end of the frame for them to do what they wish (Render to Profiler Stats Overlay or Display/Record in a Profiler or file)
+
+As this data is largely a struct like object, accessing its data will be as simple as accessing fields. Additionally the rendering or logging of these statistics will fall on the user and not affect the Tick or be done at all within the MLAPI framework.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+
+In order to limit the interface, we are pushing the collected set or performance data to be obtainable only at the end of a tick. The following is roughly what this object could look like. Currently this object will be updated in ```NetworkManager.LateUpdate()``` for the MLAPI and ```Transport.Send()``` and other relevant transport method for the transport layer. At the end of the tick the user will be notified that the relevant data is complete and ready (both the MLAPI data and Transport data).
```
becksebenius-unity, line 32: I mentioned this in other comments but just wanted to clearly state, I would like to know more about _how_ the data is fulfilled by these methods. How does the transport get access to the profiler data to update? If they're passed by reference (`struct`) how is the data consolidated?

text/0003-performance-data.md
kvassall-unity: I have it here as means to allow people to publicly comment on its possibility. I rather have it here with overwhelming sentiments of its removal then to have a large amount of people wonder why the implementation was so closed.

text/0003-performance-data.md
MFatihMAR: since it's a relatively small container, we could make it a `struct` and forget about heap pressure.
on the other hand, we could leave it as a `class`, cache it internally on the producer-side (transport in this case) and re-use that existing reference to modify properties everytime we call transport's (let's say) `TickProfilerStats()` API.

text/0003-performance-data.md
MFatihMAR: I think this approach it mixing 2 other distinct approaches:

- be opinionated: type types into struct and pass it though
- be unopinionated: fill Dict with keys-values and pass it through

In my opinion, we should settle on one approach (either opinionated or unopinionated struct passthrough) and don't mix both of them.

In this scenario, I'm more inclined towards being unopinionated and just returning a Dict<string, int> back but that comes with its downsides:

- Dict<string, int> could be too limited — maybe everything can't be ints?
- Dict<string, object> introduces boxing overhead

I'd like to discuss more about this struct's layout, let me know what you guys think :)

text/0003-performance-data.md
```diff
@@ -0,0 +1,83 @@
...
+In order to allow first and third party rendering of the data we will need to allow both MLAPI and its transports to propagate data up the chain. 
+If the profiler flags are enabled the following will need to happen: 
+Transports will need to output the number of bytes being sent in the current tick and the latency. The data will be written out to a ProfilerTickData chunk. This data will be generic and transport agnostic.
+MLAPI will need to output the number of networked objects and the amount of RPCs being sent. This data will be added to the ProfilerTickData chunk.
+
+External Tools and APIs will be able to then get this fully written ProfilerTickData at the end of the frame for them to do what they wish (Render to Profiler Stats Overlay or Display/Record in a Profiler or file)
+
+As this data is largely a struct like object, accessing its data will be as simple as accessing fields. Additionally the rendering or logging of these statistics will fall on the user and not affect the Tick or be done at all within the MLAPI framework.
+
+# Reference-level explanation
```
MFatihMAR, line 28: I'd like to see a code-snippet that shows required changes on MLAPI's Transport interface.

More specifically in this file/class:

> Runtime/Transports/Transport.cs

```cs
namespace MLAPI.Transports
{
    /// <summary>
    /// A network transport
    /// </summary>
    public abstract class Transport : MonoBehaviour
    {
        // ...
```

text/0003-performance-data.md
```diff
@@ -0,0 +1,83 @@
...
+[summary]: #summary
+
+Currently there is Transport performance data and MLAPI performance data that is not being propagated up externally for any consumer to render or consume. The use cases for this would be useful for technologies such as a **Network Stats Overlay** or a **Profiler** (**Network Messages** and **Network Operation** need this data in the vanilla Unity Profiler). Users need a viable and low overhead way to obtain this data (such as RPCs sent this frame/tick) from MLAPI and utilize this data in the way that they want for their own metrics.
+
+# Motivation
+[motivation]: #motivation
+
+This is needed to provide both a default and user specific implementations for future Network Profilers and Network Stats overlays. The MLAPI source code can remain Open Source, and yet, the data being streamed can be hooked up to external and low level systems.
+
+# Guide-level explanation
```
MFatihMAR, line 16: (optionally)
Adding some fake or real example use-cases under this section would be helpful.
I'm thinking transport adding bytes sent etc and MLAPI adding rpcs sent etc similarly you mentioned below but with more clear "here's how we'd produce them" highlights.

text/0003-performance-data.md
kvassall-unity: If thats the line we want to make then i favor typed types into the struct. There shouldn't be that many cases we aren't capable of thinking of. Most are known and any new ones can be added.

text/0003-performance-data.md
kvassall-unity: updated

text/0003-performance-data.md
```diff
@@ -0,0 +1,133 @@
...
+    int numberOfRPCs;
+    ...
+
+    // transport-level
+    TransportProfilerData transportProfilerData;
+    ...
+}
+
+
+struct TransportProfilerData {
```
LukeStampfli, line 48: I don't understand why to retrieve this data a change to each Transport is necessary.
Bytes/send received is just the actual byte buffers send/received without all the header overhead from the transport layer right?
Same for numberOfMessagesOutgoing. That's "Number of send calls from MLAPI to the transport" and not number of networked packets sent out after the transport batched them.

So if you ask me we can move all this code to the place where Transport.Send is called and don't change transports at all.

MFatihMAR: > Bytes/send received is just the actual byte buffers send/received without all the header overhead from the transport layer right?

Not necessarily, there could be more messages going back and forth (keep-alive pings, other QoS packets etc.) which could be sent and received without explicit transport-API-level `Send()` or `Poll()`.

> Same for numberOfMessagesOutgoing. That's "Number of send calls from MLAPI to the transport" and not number of networked packets sent out after the transport batched them.

Again, not necessarily. We might end up sending more or less number of packets on the transport-side.

> So if you ask me we can move all this code to the place where Transport.Send is called and don't change transports at all.

Also, this would mean that we will ALWAYS expect that struct being filled or left unfilled but returned back on every `Send()` — what if we never call `Send()` in a transport tick frame?

Beyond that, retrieving this profile stats data struct will be over on its own API which could even be wrapped into `#if UNITY_NETWORK_PROFILER`-like macro and left out entirely from release builds.

I think we're going in the right direction overall but we should discuss and decide what that intermediary struct should look like.

text/0003-performance-data.md
```diff
@@ -0,0 +1,133 @@
...
+This design of single struct is the best for locality of data. 
+There is less room for user error since there is only one readonly object that they get access to.
+There is no room for user injected data or exceptions since we aren’t using an event signaling or actions. All user interaction for rendering happens at end of tick so they can’t affect the time of profiling or cause unforeseen issues.
+
+The impact of not doing a Network Tick and Profiler Notification System is that we’d loose the Profiling capabilities that existed in the vanilla Profiler with Mirror/UNET.
+
+# Prior art
+[prior-art]: #prior-art
+
+Writing all pertinent data per tick is a similar method to what most rendering pipelines do, which is gather relevant data store in a buffer and render it. Since Multiplayer requires the same effective speed as our Render Pipelines we want to make sure it is as stable and robust as possible.
```
MFatihMAR, line 122: in terms of profiler stats data, what Render Pipelines do in that area?

text/0003-performance-data.md
MFatihMAR: in terms of API-design, I'm still not sure `Dict<string, object>` vs `struct OurCustomProfilingData` here. Dict approach would give us more flexibility with the cost of performance and less clue what will that Dict will contain — on the other hand, custom struct approach might be too opinionated and require lots of iterative changes over time.

I think we should discuss about this in a wider group, I will schedule a meeting for us :)

text/0003-performance-data.md
LukeStampfli: I think that's t ne problem with this RFC/interface. It's not clear about what exactly we are measuring. Should the transport be doing keep alive etc.

Often the transports lacks an API to properly return amount of network packets sent and how many bytes sent.
We already have this problem with GetCurrentRTT of the transport interface. There are many ways to define RTT it's not entirely clear what exactly this means and different Transports right now have very different implementations.

If we add metrics to an interface like Transport they should be clearly defined which currently is not the case.

MFatihMAR: I also agree that we should have a better API design and a better RFC document here. Let's discuss this in a meeting together!

text/0003-performance-data.md
kvassall-unity: We're going with prenamed fields in a struct. If anything we can add a dictionary to it at some later date. but more than likely we will just add the missing fields

text/0003-performance-data.md
kvassall-unity: Read/Write to one or two buffers typically accessible by most of the pipeline (Just at different times)

kvassall-unity: Profiler stats will just have this same buffer (as a struct) and can use it to create a history or to render, or do whatever it wants with this data

text/0003-performance-data.md
kvassall-unity: Updated to be more explicit as to why MLAPI is responsible for aggregating and controlling the buffer/performance data

wackoisgod:
So my main question is why do we need this structs?  if we want to add a field we need to update the MLAPI and I am not sure why they need to be this explicit when at the end of the day its key value pairs that get displayed on a screen or passed to the Unity profiler. I also worry about if the transport wants to report specific data that may not be shared for all transport we just don't get that data. 

We also get into this case where this struct just ends up growing forever because we can't remove or change fields in the future for fear of breaking the API contract we have with already released packages that depend on MLAPI. 

I also know @LukeStampfli brought this up but I don't get why we aren't just allowing the Transport to subscribe to some callback from the profiler system instead of changing the Transport API. 

I would much rather see in the Transport: 

```
void Init(){
  ProfilerSystem.OnGetProfilerData += GetProfilerData;
}

List<ProfilerCounterData> GetProfilerData() {
   return data;
}
```

kvassall-unity:
> So my main question is why do we need this structs? if we want to add a field we need to update the MLAPI and I am not sure why they need to be this explicit when at the end of the day its key value pairs that get displayed on a screen or passed to the Unity profiler. I also worry about if the transport wants to report specific data that may not be shared for all transport we just don't get that data.
> 
> We also get into this case where this struct just ends up growing forever because we can't remove or change fields in the future for fear of breaking the API contract we have with already released packages that depend on MLAPI.
> 
> I also know @LukeStampfli brought this up but I don't get why we aren't just allowing the Transport to subscribe to some callback from the profiler system instead of changing the Transport API.
> 
> I would much rather see in the Transport:
> 
> ```
> void Init(){
>   ProfilerSystem.OnGetProfilerData += GetProfilerData;
> }
> 
> List<ProfilerCounterData> GetProfilerData() {
>    return data;
> }
> ```

I had it as a struct with a  dictionary for unknown cases. Then we went struct. Now dictionary. I have no fight in which method. Whomever is closer to the transport should just make the call here. Its not big issue to me (performance or API wise)

The only difference between the callback and direct call is the expectation that every transport implements (or is aware) the contract. As a callback the transport may never implement the event. I'm fine with switching it.

wackoisgod:
> > So my main question is why do we need this structs? if we want to add a field we need to update the MLAPI and I am not sure why they need to be this explicit when at the end of the day its key value pairs that get displayed on a screen or passed to the Unity profiler. I also worry about if the transport wants to report specific data that may not be shared for all transport we just don't get that data.
> > We also get into this case where this struct just ends up growing forever because we can't remove or change fields in the future for fear of breaking the API contract we have with already released packages that depend on MLAPI.
> > I also know @LukeStampfli brought this up but I don't get why we aren't just allowing the Transport to subscribe to some callback from the profiler system instead of changing the Transport API.
> > I would much rather see in the Transport:
> > ```
> > void Init(){
> >   ProfilerSystem.OnGetProfilerData += GetProfilerData;
> > }
> > 
> > List<ProfilerCounterData> GetProfilerData() {
> >    return data;
> > }
> > ```
> 
> I had it as a struct with a dictionary for unknown cases. Then we went struct. Now dictionary. I have no fight in which method. Whomever is closer to the transport should just make the call here. Its not big issue to me (performance or API wise)
> 
> The only difference between the callback and direct call is the expectation that every transport implements (or is aware) the contract. As a callback the transport may never implement the event. I'm fine with switching it.

I guess I am not clear on why we don't just use the dictionary in all cases ? 

As for the callback, I mean there is nothing requiring the transport for filling out the struct either. I just worry about breaking the transport or adding something to it when we are wanting to change it and if the structs are needed.

kvassall-unity:
> > > So my main question is why do we need this structs? if we want to add a field we need to update the MLAPI and I am not sure why they need to be this explicit when at the end of the day its key value pairs that get displayed on a screen or passed to the Unity profiler. I also worry about if the transport wants to report specific data that may not be shared for all transport we just don't get that data.
> > > We also get into this case where this struct just ends up growing forever because we can't remove or change fields in the future for fear of breaking the API contract we have with already released packages that depend on MLAPI.
> > > I also know @LukeStampfli brought this up but I don't get why we aren't just allowing the Transport to subscribe to some callback from the profiler system instead of changing the Transport API.
> > > I would much rather see in the Transport:
> > > ```
> > > void Init(){
> > >   ProfilerSystem.OnGetProfilerData += GetProfilerData;
> > > }
> > > 
> > > List<ProfilerCounterData> GetProfilerData() {
> > >    return data;
> > > }
> > > ```
> > 
> > 
> > I had it as a struct with a dictionary for unknown cases. Then we went struct. Now dictionary. I have no fight in which method. Whomever is closer to the transport should just make the call here. Its not big issue to me (performance or API wise)
> > The only difference between the callback and direct call is the expectation that every transport implements (or is aware) the contract. As a callback the transport may never implement the event. I'm fine with switching it.
> 
> I guess I am not clear on why we don't just use the dictionary in all cases ?
> 
> As for the callback, I mean there is nothing requiring the transport for filling out the struct either. I just worry about breaking the transport or adding something to it when we are wanting to change it and if the structs are needed.

Switching to Dictionary and callback from transport. 

In regards to the Dictionary, the concern was boxing and performance but i think this is mostly negligible. I'll refactor to Dict<Enum,int> which should handle 90% of cases if not all

becksebenius-unity:
> > > > So my main question is why do we need this structs? if we want to add a field we need to update the MLAPI and I am not sure why they need to be this explicit when at the end of the day its key value pairs that get displayed on a screen or passed to the Unity profiler. I also worry about if the transport wants to report specific data that may not be shared for all transport we just don't get that data.
> > > > We also get into this case where this struct just ends up growing forever because we can't remove or change fields in the future for fear of breaking the API contract we have with already released packages that depend on MLAPI.
> > > > I also know @LukeStampfli brought this up but I don't get why we aren't just allowing the Transport to subscribe to some callback from the profiler system instead of changing the Transport API.
> > > > I would much rather see in the Transport:
> > > > ```
> > > > void Init(){
> > > >   ProfilerSystem.OnGetProfilerData += GetProfilerData;
> > > > }
> > > > 
> > > > List<ProfilerCounterData> GetProfilerData() {
> > > >    return data;
> > > > }
> > > > ```
> > > 
> > > 
> > > I had it as a struct with a dictionary for unknown cases. Then we went struct. Now dictionary. I have no fight in which method. Whomever is closer to the transport should just make the call here. Its not big issue to me (performance or API wise)
> > > The only difference between the callback and direct call is the expectation that every transport implements (or is aware) the contract. As a callback the transport may never implement the event. I'm fine with switching it.
> > 
> > 
> > I guess I am not clear on why we don't just use the dictionary in all cases ?
> > As for the callback, I mean there is nothing requiring the transport for filling out the struct either. I just worry about breaking the transport or adding something to it when we are wanting to change it and if the structs are needed.
> 
> Switching to Dictionary and callback from transport.
> 
> In regards to the Dictionary, the concern was boxing and performance but i think this is mostly negligible. I'll refactor to Dict<Enum,int> which should handle 90% of cases if not all

Suggestion for the dictionary to use string keys instead of a predefined enum. unintuitively, it's actually better for performance (enums as dictionary keys actually get boxed 🤢). and if we're going to go with a flexible approach, lets maximize that instead of restricting them to predefined keys, that way users building their own tools for their own custom transport could extend. we can still provide predefined keys for the user in a constants class so that we can guarantee good data for our profiler.

becksebenius-unity:
Not sure if this is the callback pattern you're referring to but want to comment on this snippet that Andrew posted

`void Init(){
  ProfilerSystem.OnGetProfilerData += GetProfilerData;
}

List<ProfilerCounterData> GetProfilerData() {
   return data;
}`

this seems like a problematic api design. return values on multicast delegates is basically nonsensical. (i *think* c# lets you do it, but only returns the result of one of the delegates). we should either correctly support multiple return values or limit it to only one registered callback.

that said, i'm also not really a big fan of this approach. from an api design perspective it's a very indirect way of providing profiler data. if we're looking for an opt-in approach, i'd recommend that we add an interface that the transport can optionally implement, and MLAPI will try-cast the transport to that interface to see if it can gather profiler data.

wackoisgod:
> Not sure if this is the callback pattern you're referring to but want to comment on this snippet that Andrew posted
> 
> ```
> void Init(){
>   ProfilerSystem.OnGetProfilerData += GetProfilerData;
> }
> 
> List<ProfilerCounterData> GetProfilerData() {
>    return data;
> }
> ```
> 
> this seems like a problematic api design. return values on multicast delegates is basically nonsensical. (i _think_ c# lets you do it, but only returns the result of one of the delegates). we should either correctly support multiple return values or limit it to only one registered callback.
> 
> that said, i'm also not really a big fan of this approach. from an api design perspective it's a very indirect way of providing profiler data. if we're looking for an opt-in approach, i'd recommend that we add an interface that the transport can optionally implement, and MLAPI will try-cast the transport to that interface to see if it can gather profiler data.

This was just a suggestion :D I would not be against a Interface and that may indeed allow for the most flexibility without breaking anything.

kvassall-unity:
> > > > > So my main question is why do we need this structs? if we want to add a field we need to update the MLAPI and I am not sure why they need to be this explicit when at the end of the day its key value pairs that get displayed on a screen or passed to the Unity profiler. I also worry about if the transport wants to report specific data that may not be shared for all transport we just don't get that data.
> > > > > We also get into this case where this struct just ends up growing forever because we can't remove or change fields in the future for fear of breaking the API contract we have with already released packages that depend on MLAPI.
> > > > > I also know @LukeStampfli brought this up but I don't get why we aren't just allowing the Transport to subscribe to some callback from the profiler system instead of changing the Transport API.
> > > > > I would much rather see in the Transport:
> > > > > ```
> > > > > void Init(){
> > > > >   ProfilerSystem.OnGetProfilerData += GetProfilerData;
> > > > > }
> > > > > 
> > > > > List<ProfilerCounterData> GetProfilerData() {
> > > > >    return data;
> > > > > }
> > > > > ```
> > > > 
> > > > 
> > > > I had it as a struct with a dictionary for unknown cases. Then we went struct. Now dictionary. I have no fight in which method. Whomever is closer to the transport should just make the call here. Its not big issue to me (performance or API wise)
> > > > The only difference between the callback and direct call is the expectation that every transport implements (or is aware) the contract. As a callback the transport may never implement the event. I'm fine with switching it.
> > > 
> > > 
> > > I guess I am not clear on why we don't just use the dictionary in all cases ?
> > > As for the callback, I mean there is nothing requiring the transport for filling out the struct either. I just worry about breaking the transport or adding something to it when we are wanting to change it and if the structs are needed.
> > 
> > 
> > Switching to Dictionary and callback from transport.
> > In regards to the Dictionary, the concern was boxing and performance but i think this is mostly negligible. I'll refactor to Dict<Enum,int> which should handle 90% of cases if not all
> 
> Suggestion for the dictionary to use string keys instead of a predefined enum. unintuitively, it's actually better for performance (enums as dictionary keys actually get boxed 🤢). and if we're going to go with a flexible approach, lets maximize that instead of restricting them to predefined keys, that way users building their own tools for their own custom transport could extend. we can still provide predefined keys for the user in a constants class so that we can guarantee good data for our profiler.

Yes, By enum i meant defined string.

> > Not sure if this is the callback pattern you're referring to but want to comment on this snippet that Andrew posted
> > ```
> > void Init(){
> >   ProfilerSystem.OnGetProfilerData += GetProfilerData;
> > }
> > 
> > List<ProfilerCounterData> GetProfilerData() {
> >    return data;
> > }
> > ```
> > 
> > 
> > this seems like a problematic api design. return values on multicast delegates is basically nonsensical. (i _think_ c# lets you do it, but only returns the result of one of the delegates). we should either correctly support multiple return values or limit it to only one registered callback.
> > that said, i'm also not really a big fan of this approach. from an api design perspective it's a very indirect way of providing profiler data. if we're looking for an opt-in approach, i'd recommend that we add an interface that the transport can optionally implement, and MLAPI will try-cast the transport to that interface to see if it can gather profiler data.
> 
> This was just a suggestion :D The primary suggestion was I think it should be opt in vs a requirement.
> 
> I would not be against a Interface and that may indeed allow for the most flexibility without breaking anything.

Given that the transports simply subscribe and provide the data to MLAPI as opt in. I see no reason for an interface. If the ```GetData()``` actually held logic id be more prone to go this way. Currently i don't see much of a use it will have except for adding type casting code. Callbacks seems better to me since we're talking about data and not logic.


Updated everything to the latest and greatest feedback

wackoisgod:
I think @becksebenius-unity concern around delegates in C# is a valid one; they are kind of trash especially the multi-cast ones. They are especially gross in IL2CPP :( where as the interface cost will be cheaper for those platforms.

text/0003-performance-data.md
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	public class ProfilerTickData {
+		public int tickID;
+
+		public readonly Dictionary<string, int> tickData = new Dictionary<string, int>();
+	}
+
+	public interface ITransportProfilerData {
+		Dictionary<string, int> GetTransportGetData();
```
becksebenius-unity, line 55: This is an implementation detail and IMO doesn't block this RFC, but we should pass in the dictionary rather than having it return it so that we can re-use the dictionary between network ticks and avoid memory allocs

becksebenius-unity: I see below that the unet transport example does cache its own dictionary. this approach could work as well, although I think this adds complexity because the internal dictionary is used externally and the transport doesn't know when it is safe to clear the dictionary for the next network tick

kvassall-unity: Yes the transport is in control of its own memory block and deciding what goes into it. It will cache regardless.
And yes i'm willing to refactor here in an actual PR here for performance gains. Passing in a dictionary to the transport might be useful in terms of clarity even though it will still be caching locally.

MFatihMAR: I'd like to add a similar implementation-detail question here — it might be too constraining to map `string` to `int` but not `string` to `object`. Since performance won't be a big concern here (throwing boxing cost out of the window), it might be useful (and more future-proof) to map `string` to `object` and allow passing other types such as `float` etc through this proposed design.

kvassall-unity: I only somewhat agree. But for something that is typically a float (like timestamp) we can just represent with milliseconds. And outside of that i failed to come up with an example that required floats or a vector (which you can just split it up into three ints) 

Having a contrarian example (which i don't have) would help me take a stronger position one way or the other

becksebenius-unity (approved):

text/0007-performance-data-propagation.md
```diff
@@ -0,0 +1,179 @@
...
+In order to limit the interface, we are pushing the collected set or performance data to be obtainable only at the end of a tick. The following is roughly what this object could look like. Currently this object will be updated in `NetworkManager.LateUpdate()` for the MLAPI and `Transport.Send()` and other relevant transport method for the transport layer. At the end of the tick the user will be notified that the relevant data is complete and ready (both the MLAPI data and Transport data). Since MLAPI is responsible for collecting the data, and the data from transports will be generic and apply to all/most transports, MLAPI as  the owner of the collection of performance data will call `GetTransportGetData()` on the transports through an interface in order to get the data exactly when it needs it in the pipeline before propagating the data out externally. Whether or not the data is collected per transport can also be toggled by a runtime flag `profilerEnabled`.
+
+## Pseudo-code
+
+```cs
+namespace MLAPI
+{
+    class ProfilerConstants
+    {
+        public const string NumberOfServerRPCs = "numberOfServerRPCs";
```
mattwalsh-unity, line 43: I recently went through an exercise to clean up the channel stuff that might be of use here.  Check out Transport.cs circa line 9 / the 'Channel' definition.  You can define each measurement as an enum, then if you need to output a string you can just call ToString() on it.  Then you don't have to maintain the strings / pay for the string de-ref.

mattwalsh-unity (commented):
Added a comment

text/0007-performance-data-propagation.md
```diff
@@ -0,0 +1,179 @@
...
+
+There is less room for user error since there is only one readonly object that they get access to.
+
+There is no room for user injected data or exceptions since we aren’t using an event signaling or actions. All user interaction for rendering happens at end of tick so they can’t affect the time of profiling or cause unforeseen issues.
+
+The impact of not doing a Network Tick and Profiler Notification System is that we’d loose the Profiling capabilities that existed in the vanilla Profiler with Mirror/UNET.
+
+# Prior art
+[prior-art]: #prior-art
+
```
mattwalsh-unity, line 168: There is already a stat gathering feature added to MLAPI that might be extendable / have things we can steal (see ProfilerStat / ProfilerStatManager).  It does a moving average.  In particular I had to put in some smarts in the 'Record' method to handle when tons of samples are taken in a short period of time.

text/0007-performance-data-propagation.md
```diff
@@ -0,0 +1,179 @@
...
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+In order to allow first and third party rendering of the data we will need to allow both MLAPI and its transports to propagate data up the chain. 
+
+If the profiler flags are enabled the following will need to happen:
+
+Transports will need to output the number of bytes being sent in the current tick and the latency. The data will be written out to a `ProfilerTickData` chunk. This data will be generic and transport agnostic.
+
+MLAPI will need to output the number of networked objects and the amount of RPCs being sent. This data will be added to the `ProfilerTickData` chunk.
```
mattwalsh-unity, line 25: As in the # of networked objects that are currently being managed?  Subtle point, would the AOI-ness matter?  Like would we count objects that are outside AOI for a client?

text/0003-performance-data.md
MFatihMAR: I think I'm still 51% on `string`—`int` mapping side but wanted to call out just in case. I'll try to convince both myself and you as well if I could find better examples.

text/0007-performance-data-propagation.md
```diff
@@ -0,0 +1,179 @@
...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+In order to limit the interface, we are pushing the collected set or performance data to be obtainable only at the end of a tick. The following is roughly what this object could look like. Currently this object will be updated in `NetworkManager.LateUpdate()` for the MLAPI and `Transport.Send()` and other relevant transport method for the transport layer. At the end of the tick the user will be notified that the relevant data is complete and ready (both the MLAPI data and Transport data). Since MLAPI is responsible for collecting the data, and the data from transports will be generic and apply to all/most transports, MLAPI as  the owner of the collection of performance data will call `GetTransportGetData()` on the transports through an interface in order to get the data exactly when it needs it in the pipeline before propagating the data out externally. Whether or not the data is collected per transport can also be toggled by a runtime flag `profilerEnabled`.
```
mattwalsh-unity, line 34: Does this behavior vary whether we're a client or server?

And if we're a server, is this data aggregated for all connected clients or separated out per client?  For example, with AOI the amount of data replicated will vary by client connection.  Would you want to be able to see that?  Or is it enough to lump it into one pile for all connections/clients?

text/0007-performance-data-propagation.md
```diff
@@ -0,0 +1,179 @@
...
+        private static int tickID = 0;
+
+        public delegate void PerfomanceDataEventHandler(ProfilerTickData profilerData);
+
+        event PerfomanceDataEventHandler PerfomanceDataEvent;
+
+        void LateUpdate()
+        {
+            // ...
+            ProfilerTickData profilerData = new ProfilerTickData();
```
mattwalsh-unity, line 79: Probably should use a pool, (impl I know)

text/0007-performance-data-propagation.md
kvassall-unity: From another comment by @becksebenius-unity :

"Suggestion for the dictionary to use string keys instead of a predefined enum. unintuitively, it's actually better for performance (enums as dictionary keys actually get boxed 🤢). and if we're going to go with a flexible approach, lets maximize that instead of restricting them to predefined keys, that way users building their own tools for their own custom transport could extend. we can still provide predefined keys for the user in a constants class so that we can guarantee good data for our profiler."

text/0007-performance-data-propagation.md
kvassall-unity: That code will still be used. This should feed into this for the Profiler among other things.

text/0007-performance-data-propagation.md
kvassall-unity: Thats an example. but ultimately it would be up to how that data is captured by MLAPI and isn't dependent on whether or not we have this system.

text/0007-performance-data-propagation.md
kvassall-unity: This is dependent on what MLAPI wants to output. Currently i'll match 1-to-1 with existing Profiling which is **all connections** and **server** based.

text/0007-performance-data-propagation.md
kvassall-unity: Will likely use a static pool where the receiver (third party) can't keep a reference or the reference is readonly

text/0007-performance-data-propagation.md
MFatihMAR: I think he's not suggesting to change `Dictionary<string, int>` but to define an enum for "our" `ProfilerConstants` to be an `enum` and do `ProfilerConstants. NumberOfServerRPCs.ToString()` instead of string constants.

kvassall-unity: Ah i see. I can update to that. in the actual PR

text/0007-performance-data-propagation.md
becksebenius-unity: I interpreted this more as, how many objects sent state updates in this network tick. Both might be helpful though

mattwalsh-unity (approved):

wackoisgod (approved):

PR #8, opened 2021-01-31
Network Update Loop

MFatihMAR:
[See rendered RFC document &rarr;](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/network-ticker/text/0008-network-update-loop.md)


NoelStephensUnity (approved):
Thank you Fatih!
Looks very nice and a better implementation!

LukeStampfli (approved):
Clean and decoupled API, I like it!

text/0000-network-update-loop.md
```diff
@@ -0,0 +1,720 @@
+- Feature Name: `network-update-loop`
+- Start Date: 2021-01-31
+- RFC PR: [Unity-Technologies/com.unity.multiplayer.rfcs#0000](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/pull/0000)
+- Issue: [Unity-Technologies/com.unity.multiplayer#0000](https://github.com/Unity-Technologies/com.unity.multiplayer/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Often there is a need to update netcode systems like RPC queue, transport, and others outside the standard [`MonoBehaviour` event cycle](https://docs.unity3d.com/Manual/ExecutionOrder.html).
```
becksebenius-unity, line 9: Could you provide some example use cases? I'm not sure I understand why a user needs this.

MFatihMAR: Network systems will be implemented on top of this infrastructure, some examples are RPC queue, transport IO and other potentials are also outlined in the "Future possibilities" at the bottom of the document.

Framework users may or may not utilize this skeleton — we could expect users to tie their game logic into some network systems running at the same stage.

text/0000-network-update-loop.md
```diff
@@ -0,0 +1,712 @@
...
+    NetworkUpdateLoop -> NetworkUpdateLoop: ++FrameCount
+    NetworkUpdateLoop -> NetworkUpdateLoop: UpdateStage = Initialization
+    loop m_Initialization_Array
+        NetworkUpdateLoop -> INetworkUpdateSystem: NetworkUpdate
+        NetworkUpdateLoop <-- INetworkUpdateSystem
+    end
+    PlayerLoop <-- NetworkUpdateLoop
+end
+group EarlyUpdate
+    PlayerLoop -> NetworkUpdateLoop: RunNetworkEarlyUpdate
```
JesseOlmer, line 251: This is probably too early - e.g. before profilerstartframe definitely seems wrong... maybe tie it closer to other networking needs (UnityWebRequestUpdate)

text/0000-network-update-loop.md
```diff
@@ -0,0 +1,712 @@
...
+    PlayerLoop -> PlayerLoop: UpdateKinect
+    PlayerLoop -> PlayerLoop: DeliverIosPlatformEvents
+    PlayerLoop -> PlayerLoop: TangoUpdate
+    PlayerLoop -> PlayerLoop: DispatchEventQueueEvents
+    PlayerLoop -> PlayerLoop: PhysicsResetInterpolatedTransformPosition
+    PlayerLoop -> PlayerLoop: SpriteAtlasManagerUpdate
+    PlayerLoop -> PlayerLoop: PerformanceAnalyticsUpdate
+end
+group FixedUpdate
+    PlayerLoop -> NetworkUpdateLoop: RunNetworkFixedUpdate
```
JesseOlmer, line 290: Why not closer to `ScriptRunBehaviourFixedUpdate`?

text/0000-network-update-loop.md
```diff
@@ -0,0 +1,712 @@
...
+        NetworkUpdateLoop <-- INetworkUpdateSystem
+    end
+    PlayerLoop <-- NetworkUpdateLoop
+    PlayerLoop -> PlayerLoop: ScriptRunBehaviourUpdate
+    PlayerLoop -> PlayerLoop: ScriptRunDelayedDynamicFrameRate
+    PlayerLoop -> PlayerLoop: ScriptRunDelayedTasks
+    PlayerLoop -> PlayerLoop: DirectorUpdate
+end
+group PreLateUpdate
+    PlayerLoop -> NetworkUpdateLoop: RunNetworkPreLateUpdate
```
JesseOlmer, line 343: Again, maybe better served closer to `ScriptRunBehaviourLateUpdate`

text/0000-network-update-loop.md
```diff
@@ -0,0 +1,712 @@
...
+    PlayerLoop -> PlayerLoop: TriggerEndOfFrameCallbacks
+    PlayerLoop -> PlayerLoop: GUIClearEvents
+    PlayerLoop -> PlayerLoop: ShaderHandleErrors
+    PlayerLoop -> PlayerLoop: ResetInputAxis
+    PlayerLoop -> PlayerLoop: ThreadedLoadingDebug
+    PlayerLoop -> PlayerLoop: ProfilerSynchronizeStats
+    PlayerLoop -> PlayerLoop: MemoryFrameMaintenance
+    PlayerLoop -> PlayerLoop: ExecuteGameCenterCallbacks
+    PlayerLoop -> PlayerLoop: ProfilerEndFrame
+    PlayerLoop -> NetworkUpdateLoop: RunNetworkPostLateUpdate
```
JesseOlmer, line 400: This is too late - after profilerendframe, after memory frame maintenance, after a bunch of graphics internals...

text/0000-network-update-loop.md
```diff
@@ -0,0 +1,712 @@
...
+  - unregisters an `INetworkUpdateSystem` from all `NetworkUpdateStage`s
+
+### Frames and Stages
+
+After injection, player loop will look like this:
+
+- **Initialization**
+  - // other systems
+  - RunNetworkInitialization
+    - `NetworkLoopSystem.AdvanceFrame()`
```
JesseOlmer, line 51: Feels like frame id would be an INetworkUpdateSystem itself, not a core part of the base loop? Why would we want to match the network frame count to the runtime frame count and not to, e.g., when networking is initialized? Maybe I'm reading too much into the psuedo-code here?

text/0000-network-update-loop.md
```diff
@@ -0,0 +1,712 @@
...
+end
+```
+</details>
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+N/A: We need to have this, other systems such as network tick, network variable snapshotting and others will be relying on this pipeline.
+
+# Rationale and alternatives
```
JesseOlmer, line 679: I'm a bit dubious about instituting a structured loop concept that lives on top of the PlayerLoopSystem. It feels like, if executed poorly, it could limit flexibility in the PlayerLoopsystem. Maybe it's OKAY because enough other parts of the engine break if you customize the system beyond where we're hooking in, but it feels wrong that the entire system might stop working if you change your player loop enough for special case without giving the user any way to remedy it by composing their own network loop.

JesseOlmer: Perhaps an alternative approach is to model this as a series of helpers/extensions, along with some default settings, rather than a system-unto-itself.

text/0000-network-update-loop.md
MFatihMAR: no, I think you are right. frame concept should be a system on its own and not a part of the network update loop.

text/0000-network-update-loop.md
MFatihMAR: we're modifying the **existing** player loop system, we are not building a new one from scratch. users are still able to modify the player loop system after we inject our methods into the existing one — they might even unplug our systems! I believe our approach is right here but I will reconsider where we inject network methods as you suggested above.

MFatihMAR: looking at this one more time again, I think I got what you were saying. however, I still think "we" should decide where we would inject our network loop methods because it will also have implications on the framework itself. I believe I went a little bit too far without considering and fully understanding what are the steps the player loop takes (injecting after MemoryFrameMaintenance doesn't sound right). I will investigate more and elaborate reasons why certain places are selected to inject network loop methods. awesome feedback, thank you very much Jesse :)

text/0000-network-update-loop.md
```diff
@@ -0,0 +1,706 @@
...
+    {
+        // extension method calls are equivalent to `NetworkUpdateLoop.RegisterNetworkUpdate(this, NetworkUpdateStage.XXX)`
+        this.RegisterNetworkUpdate(NetworkUpdateStage.EarlyUpdate);
+        this.RegisterNetworkUpdate(NetworkUpdateStage.FixedUpdate);
+        this.RegisterNetworkUpdate(NetworkUpdateStage.Update);
+    }
+
+    // will be called on EarlyUpdate, FixedUpdate and Update stages as registered above
+    public void NetworkUpdate()
+    {
```
mattwalsh-unity, line 123: Even though we can safely access the stage via a singleton since there is only ever one Unity PlayerLoop at once, as a user I probably would find it more intuitive if the arg for what stage this is was passed in to the call else I’d have to google for how to read the singleton value.

Also this helps justify having a dedicated `INetworkUpdateSystem` interface :)

text/0000-network-update-loop.md
```diff
@@ -0,0 +1,706 @@
...
+    private static readonly List<INetworkUpdateSystem> m_EarlyUpdate_List = new List<INetworkUpdateSystem>();
+    private static INetworkUpdateSystem[] m_EarlyUpdate_Array = new INetworkUpdateSystem[0];
+
+    private static void RunNetworkEarlyUpdate()
+    {
+        UpdateStage = NetworkUpdateStage.EarlyUpdate;
+        int arrayLength = m_EarlyUpdate_Array.Length;
+        for (int i = 0; i < arrayLength; i++)
+        {
+            m_EarlyUpdate_Array[i].NetworkUpdate();
```
mattwalsh-unity, line 505: I know this is pseudocode and the RFC does not specify the underlying implementation, yet thought I'd suggest rather than have individually named variables with and duplicated loop code to invoke it to rather have something like a map of the enums to arrays.  Then you could just have one function for 'Run****' which takes the enum and does the right thing.

text/0000-network-update-loop.md
mattwalsh-unity: I'm trying to unpack 'you' and 'your' in @JesseOlmer's comment:

>  it feels wrong that the entire system might stop working if you change your player loop enough for special case without giving the user any way to remedy it by composing their own network loop.

Do you mean the system will stop working if Unity makes changes to PlayerLoop?  Or if the SDK team makes changes that break existing users? Or (my guess) that users that are already aware and using Player Loop now have to sidestep this system?

text/0000-network-update-loop.md
MFatihMAR: that is less performant than this approach. I think I will plug some benchmarks into this RFC comparing list, array, hashset, dictionary containers in different places.

MFatihMAR: actually, I'd like to keep these implementation-dependent details to MLAPI implementation PR. I will be comparing, explaining, and discovering trade-offs over there.

JesseOlmer (approved):
lgtm

text/0000-network-update-loop.md
NoelStephensUnity: If I understand the concern here, maybe we offer a way for users to adjust where the Network Update Loop registers its hooks?  Perhaps this could be a new MLAPI tool within the editor that allows one to visualize where the network update loop is registered and make any adjustments relative to their own registered player loop systems?  This provides the user with the ability to make adjustments if needed, and of course the tool would have to have a way for the user to "reset back to default settings" in the event they fiddle with it too much.

PR #12, opened 2021-02-12
Synced Networked Variables

jeffreyrainy:
[See linked RFC document →](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/sync-net-vars/text/0000-sync-networked-vars.md)

RFC for the synced NetworkedVariables


text/0000-sync-networked-vars.md
```diff
@@ -0,0 +1,111 @@
...
+We suggest two elements:
+- Changing `NetworkedBehaviour::NetworkedBehaviourUpdate()` so that the function runs at most one per Network Tick. When called more than once per Network Tick, it would simply return.
+- NetworkedVariables carrying two pieces of information with their delta updates:
+    - The local tick at which a variable was modified
+    - The remote tick at which the variable was modified, when the host relays the variable to clients, including the source client    
+
+# Motivation
+[motivation]: #motivation
+
+In a client-server topology, any client and the server can modify NetworkedVariables. The server receives the clients' changes and send them back to all clients. In order to retain generality, we'll assume that any variable can be modified by anyone and that the server can also modify the variable as they are received and sent. For example, we could have server-driven physics over client-driven character movement.
```
NoelStephensUnity, line 20: minor typos to adjust:
_From this: The server receives the clients' changes and send them back to all clients_
**To This:** The server receives the clients' changes and sends them back to all of the clients

text/0000-sync-networked-vars.md
```diff
@@ -0,0 +1,111 @@
...
+
+We would like the game code on the client machine to be able to pick any interpolation system it desires for character movement. We'd like the game code to be able to pick client- or server-authoritative mechanics as it see fit. In order to allow this, we believe that each NetworkedVariable should know:
```
NoelStephensUnity, line 22: typo adjustment:
_From this: We would like the game code on the client machine to be able to pick any interpolation system it desires for character movement.  We'd like the game code to be able to pick client- or server-authoritative mechanics as it see fit.  In order to allow this, we believe that each NetworkedVariable should know:
- When was I modified by a client?
- When did this update travel through the server?
**- Do I have more recent local information about my value?**
_
**To this:** It is proposed that the NetworkedVariable  interpolation algorithm be easily customizable and that the user can chose whether it supports client or server side authoritative  mechanics.  To account for this, each NetworkedVaraible will keep track of the following:
- When it was modified
- When the server updated it
**- Do I have more recent local information about my value?**
(not 100% sure what the last line item implies)

text/0000-sync-networked-vars.md
```diff
@@ -0,0 +1,111 @@
...
+- When was I modified by a client?
+- When did this update travel through the server?
+- Do I have more recent local information about my value?
+
+This will enable better client-side interpolation and lag compensation, in the future, and provide flexibility for systems like snapshotting. 
+
+This will enable the client to tell the difference between receiving its own updates back and receiving updates back with server-side adjustments for things like physics.
+
+This will keep the overall bandwidth usage in check, preventing frequent updates for NetworkedVariables.
```
NoelStephensUnity, line 31: _From: This will enable better client-side interpolation and lag compensation, in the future, and provide flexibility for systems like snapshotting._

**To:** This will enable better client-side interpolation, lag compensation, and in the future provide the required flexibility for new systems (i.e. like snapshotting).

_From: This will enable the client to tell the difference between receiving its own updates back and receiving updates back with server-side adjustments for things like physics._

**To:** This provides enough information such that the client can distinguish between its own updates from that of server-side adjustments (i.e. like physics).

text/0000-sync-networked-vars.md
```diff
@@ -0,0 +1,110 @@
...
+
+#### Local tick
+The tick number at which a variable was modified, relative to the tick system on this machine. If I get an update for a variable I changed, this tell me: when did I change it?
+
+#### Remote tick
+The tick number at which a variable was relayed, relative to the tick system on the remote machine. If I get an update from the server, this tells me the local tick on the server when it sent the update.
+
+#### Network Tick wrapping 
+
+The tick value sent will be a rolling value, probably 8- or 16-bit. This will be hidden internally, but will keep the bandwidth usage in check.
```
MFatihMAR, line 55: 8bit (255) could be a little too small.

text/0000-sync-networked-vars.md
```diff
@@ -0,0 +1,110 @@
...
+ - `SendTickrate` and `LastSyncedTime` to be removed from `NetworkedVar`
+
+Below diagram shows a couple NetworkedVariable updates with example tick numbers.
+
+![](0000-sync-networked-vars/netvars.png)
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This increases slightly the bandwidth usage of _individual_ NetworkedVariable update it also renders the NetworkedVariables system dependent on the tick system.
```
MFatihMAR, line 84: A more generalized/global ticking & snapshotting could help in this area greatly but we're not there yet...

text/0000-sync-networked-vars.md
```diff
@@ -0,0 +1,110 @@
...
+
+# Prior art
+[prior-art]: #prior-art
+
+This is a pretty typical setup seen in other proprietary network engines.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+Do we require the tick to advance at the same rate on the client and server machine. Can we support an asymmetry in tick rate between client and server? Is this important?
```
MFatihMAR, line 99: Yes: client input sendrate vs server snapshot sendrate.
Yes: this is important :)

text/0000-sync-networked-vars.md
```diff
@@ -0,0 +1,110 @@
...
+
+Different clients will have a different relative ticks. When the host and client receive the NetworkedVariables, they can flip the local and remote tick. As such, if Alice sends an update to the Host and gets her value back, the local tick is valid for her. This update would not be valid for a different client, though. As such, when the snapshot and interpolation systems are fleshed out, a fix might be needed. It will probably involve marking the NetworkedVariables with a ClientId or client index. We prefer to leave this detail out of this RFC, for now.
```
MFatihMAR, line 101: As long as we know who owns those tick numbers, we can still workaround it. I'm taking control over an object, it's perfectly fine to throw the latest values away and start relying on my own ticks, values etc. Maybe a smooth interpolation/transition between the latest not-in-control state to the current now-in-control state be needed but that wouldn't dramatically affect the implementation IMHO.

text/0000-sync-networked-vars.md
```diff
@@ -0,0 +1,110 @@
...
+- RFC PR: [Unity-Technologies/com.unity.multiplayer.rfcs#0000](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/pull/0000)
+- Issue: [Unity-Technologies/com.unity.multiplayer#0000](https://github.com/Unity-Technologies/com.unity.multiplayer/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Make the NetworkedVariables synchronized with the tick system.
+
+We suggest two elements:
+- Changing `NetworkedBehaviour::NetworkedBehaviourUpdate()` so that the function runs at most one per Network Tick. When called more than once per Network Tick, it would simply return.
```
MFatihMAR, line 12: I think this part could be further explained: why we're making this change? etc.

text/0000-sync-networked-vars.md
```diff
@@ -0,0 +1,110 @@
...
+- `NetworkedVar` to include two new properties:
+
+      public ushort LocalTick { get; internal set; }
+      public ushort RemoteTick { get; internal set; }
+      
+ - `SendTickrate` and `LastSyncedTime` to be removed from `NetworkedVar`
+
+Below diagram shows a couple NetworkedVariable updates with example tick numbers.
+
+![](0000-sync-networked-vars/netvars.png)
```
MFatihMAR, line 79: how do we know if the local tick is our tick number or somebody else's? are we checking who owns `NetworkVariable` owned by `NetworkBehaviour` owned by `NetworkObject`s `OwnerId`?

text/0000-sync-networked-vars.md
MFatihMAR: perhaps, maybe outline the specific network update flow leads into network variable set (write) and get (read) in the reference-level explanation section below?

jeffreyrainy: Added diagram showing the send-side write of ticks. Receive-side will be coming shortly.

text/0000-sync-networked-vars.md
jeffreyrainy: > This update would not be valid for a different client, though. As such, when the snapshot and interpolation systems are fleshed out, a fix might be needed. It will probably involve marking the NetworkedVariables with a ClientId or client index. We prefer to leave this detail out of this RFC, for now.

I'm aiming for incremental steps. So, yeah, at this stage, we wouldn't know whether this local tick is relative to us or not.

MFatihMAR: what I was trying to say is that we _could_ know I think. the `NetworkBehaviour` that owns that `NetVar` is known therefore which `NetworkObject` owns that `NetworkBehaviour` and what's the `OwnerId` of that `NetworkObject` is known as well, which means we could basically say if this `NetVar` is living under "our" `NetworkObject`, it's ours and local tick is also our tick as well, right? :)

text/0000-sync-networked-vars.md
mattwalsh-unity: Worth mentioning that today NetworkedBehaviourUpdate runs based on a custom-configurable arbitrary time not a proper coordinated tick

text/0000-sync-networked-vars.md
```diff
@@ -0,0 +1,114 @@
...
+We suggest two elements:
+- Changing `NetworkedBehaviour::NetworkedBehaviourUpdate()` so that the function runs at most one per Network Tick. When called more than once per Network Tick, it would simply return.
+- NetworkedVariables carrying two pieces of information with their delta updates:
+    - The local tick at which a variable was modified
+    - The remote tick at which the variable was modified, when the host relays the variable to clients, including the source client    
+
+# Motivation
+[motivation]: #motivation
+
+In a client-server topology, any client and the server can modify NetworkedVariables. The server receives the clients' changes and send them back to all of the clients. In order to retain generality, we'll assume that any variable can be modified by anyone and that the server can also modify the variable as they are received and sent. For example, we could have server-driven physics over client-driven character movement.
```
mattwalsh-unity, line 20: Would it be too restrictive to make the rule like this:
1) variables belong to a behavior
2) behaviors belong to an object
3) objects are owned by at most 1 client or the server
4) hence, only the owner (either one and only one client or the server) can write changes?
5) and objects can change ownership

text/0000-sync-networked-vars.md
```diff
@@ -0,0 +1,114 @@
...
+
+It is proposed that the NetworkedVariable interpolation algorithm be easily customizable and that the user can chose whether it supports client or server side authoritative mechanics. To account for this, each NetworkedVaraible will keep track of the following:
+- When it was modified by a given client
```
mattwalsh-unity, line 23: If what I wrote above, then you'd keep track of object ownership changes on a tick

text/0000-sync-networked-vars.md
```diff
@@ -0,0 +1,114 @@
...
+
+This will keep the overall bandwidth usage in check, preventing frequent updates for NetworkedVariables.
+
+This sets up MLAPI for upcoming features like snapshot.
+
+# Dependency
+[dependency]: #dependency
+
+This feature depends on a Tick System that would provide:
+- A discrete measurement of the local time, based on an integer number of tick elapsed
```
mattwalsh-unity, line 38: How does this fit in a world where we have > 1 Networking manager?  We have been talking that each NM would have its own time.  That would make me think each NM would have its own tick system instance, and that each variable can associate with just one NM at a time.  That kind of fits, since an object can be in one scene at a time, and (in the normal case) a NM belongs to one and only one scene.  Hence, a network variable belongs to only one scene.  Does this hang together?

text/0000-sync-networked-vars.md
```diff
@@ -0,0 +1,114 @@
...
+- The tick number being updated at the very beginning of each frame and stable throughout the frame
+- A tick that is based on the real time elapsed since a fixed point in the past 
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Some definitions:
+
+#### Local tick
+The tick number at which a variable was modified, relative to the tick system on this machine. If I get an update for a variable I changed, this tell me: when did I change it?
```
mattwalsh-unity, line 48: I'm still coming around to why there are separate tick numbers for client and server vs. them all marching to the same tick number

text/0000-sync-networked-vars.md
mattwalsh-unity: Yeah, @MFatihMAR I'm thinking if we stipulate an object (and hence its behav's and its vars) always belong to one client or the server this falls out.

text/0000-sync-networked-vars.md
mattwalsh-unity: I need help becoming a believer in this :).  Seems like it will make even things like debugging harder if we're looking at server tick 123 and having to correlate it to client tick 456

mattwalsh-unity (commented):
Comments left

text/0000-sync-networked-vars.md
MFatihMAR: I think this is a design question: how granular ownership should be controlled? object vs behavior vs variable. so far, it sounds like netvar could be owned by either server or client (and even maybe both?). it sounds weird to me that a client could own an object and behaviour but not all netvars because one of them is owned/controlled by the server — but I can see a few use-cases that's why I'm fine with it.

in this context, netvar should not care which level ownership/control is applied and ask/query for that information from outside (hence my comment above suggesting to find out ownership by networkobject's ownerid lookup).

text/0000-sync-networked-vars.md
MFatihMAR: or outsource ownership management to a different place and query when needed. (ACL-like, potentially in the future)

text/0000-sync-networked-vars.md
MFatihMAR: a common example: 60 ticks per second for client to server input rate & 20 ticks per second for server to client snapshot rate.

text/0000-sync-networked-vars.md
```diff
@@ -0,0 +1,111 @@
...
+# Summary
+[summary]: #summary
+
+Make the NetworkedVariables synchronized with the tick system.
+
+We suggest two elements:
+- Changing `NetworkedBehaviour::NetworkedBehaviourUpdate()` so that the function runs at most one per Network Tick. When called more than once per Network Tick, it would simply return.
+- NetworkedVariables carrying two pieces of information with their delta updates:
+    - The local tick at which a variable was modified
+    - The remote tick at which the variable was modified, when the host relays the variable to clients, including the source client
```
xCocoDev, line 15: Would it be possible, to give an option to the end-user to don't send the local & remote ticks, if they don't really need it for a specific NetworkedVar. As I would imagine it will save-up some bandwith?

text/0000-sync-networked-vars.md
jeffreyrainy: The way I see it, we sync the netvars as a firs task/RFC, then once we have the tick system, we consider aligning the ticks when a game is joined. I basically tried taking the path of least resistance: make something that could work if the tick were offset'ed. If we decide to align them later, fine too.

text/0000-sync-networked-vars.md
jeffreyrainy: I'm going under the assumption that we'll have a single tick system. Different NM could point to the same tick system. If they're not static, in the normal case, a machine only has one Tick System, many NM pointing to it but we allow, for testing, multiple Tick Systems each being pointed to by multiple NM as-if we were running multiple instances.

text/0000-sync-networked-vars.md
jeffreyrainy: I'd prefer the "60 ticks per second client with 20 ticks per second server" situation to result in the server increasing the tick by 3, three times less often, while the client updates the tick by 1, 60 times per second. But we can hash that out in the TickSystem RFC. :-)

text/0000-sync-networked-vars.md
jeffreyrainy: What concerns me is that with the current roadmap, tick system and snapshot are pretty central. If we make the ticks optional, we may end up having to send a boolean telling "is this variable ticked" ? This would result in losing some of the gain bandwidth usage, but also increase complexity.

Cosmin-B: Hey @jeffreyrainy, i totally get what you mean, but I was thinking that maybe those variables are set to '-1' when they are not sent, and that would be similar to the flag that you mentioned, we can just check against client-side without the having to introduce another variable to server that purpose.

jeffreyrainy: Yeah. I get you. Let's keep that in mind when we implement it. I'll check which gain we can have around making them optional

text/0000-sync-networked-vars.md
mattwalsh-unity: hrm, I'm probably over-simplifying but I was thinking it would be all-or-nothing - an object is owned by a client or the server and hence all the variables go along for the ride.  @TwoTenPvP you've got lots of experience w/ customers here - are there cases worth doing work for where we want some but not all variables owned by the client within an object?

MFatihMAR:
hey @jeffreyrainy — do you think this RFC represents what's being worked on so far?
if so, do you think it's ready to be merged & get in?
if not, do you think we need to update this one and continue from here or are you planning to close this one out and start fresh on a new one?

jeffreyrainy:
> hey @jeffreyrainy — do you think this RFC represents what's being worked on so far?
> if so, do you think it's ready to be merged & get in?
> if not, do you think we need to update this one and continue from here or are you planning to close this one out and start fresh on a new one?

No, this gets closed. And new ones should be coming up for the Snapshot System (and maybe even just NetworkVariables)

Thanks for the reminder.

PR #14, opened 2021-05-05
NetworkTime

LukeStampfli:
[See linked RFC document →](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/network-time/text/0014-network-time.md)

RFC proposing a new network time and tick system.

### DIFF
--- a/text/0014-network-time.md
+++ b/text/0014-network-time.md
@@ -100,41 +100,17 @@ MLAPI sends, receives and processes data at fixed tick intervals. It is importan
 
 MLAPI exposes multiple APIs to allow developers to do so.
 
-- `NetworkManager.Singleton.NetworkTimeSystem.OnNetworkTick` can be subscribed to to run code every tick.
+- The `NetworkManager.Singleton.NetworkTickSystem.Tick` event can be subscribed to to run code every tick.
 - Code running in Update or FixedUpdate could check for the current tick using `NetworkManager.Singleton.NetworkTime.LocalTime.Tick`.
 - In the future we will expose more convenient ways to run logic during the network tick via a `NetworkFixedUpdate` in `NetworkBehaviour`.
 
-### INetworkTimeProvider
-
-While the built in `NetworkTimeSystem` provides a good starting point for a variety of games, sometimes more control over time is needed. `INetworkTimeProvider` is an interface which allows to override the root calculation of all network time. Note that this is an advanced feature, a faulty implementation of `INetworkTimeProvider` can result in a variety of issues.
-
-Here is an example implementation of a very simple custom `INetworkTimeProvider`:
-```csharp
-public class SampleNetworkTimeProvider : INetworkTimeProvider
-{
-    public bool AdvanceTime(ref NetworkTime localTime, ref NetworkTime serverTime, float deltaTime)
-    {
-        serverTime += deltaTime;
-        localTime = serverTime + 0.1f;
-        return true;
-    }
-
-    public void InitializeClient(ref NetworkTime localTime, ref NetworkTime serverTime)
-    {
-        localTime = serverTime + 0.1f;
-    }
-}
-```
-
-This `INetworkTimeProvider` is a very simple provider which increments `ServerTick` and `PredictedTick` at a consistent rate while running the `PredictedTick` 100 ms ahead of the `ServerTick`. This just acts as an example implementation and is not a recommended way to implement a `INetworkTimeProvider`
-
 ### Buffering
 
-When calculating time MLAPI adds additional buffering to delay messages further to smooth out the data stream. The default buffer values are `1f / Tickrate` when the client receives and `2f / Tickrate` when the server receives.
+When calculating time MLAPI adds additional buffering to delay messages further to smooth out the data stream. This buffering is already incorporate in the time values exposed by the `NetworkTimeSystem` The default buffer values are `1f / Tickrate` when the client receives and `1f / Tickrate` when the server receives.
 
 Buffering will only apply on NetworkVariable changes. RPCs, NetworkObject spawns etc. are unaffected.
 
-Buffer time can be manually changed in code to for instance use different buffer sizes for different platforms. `NetworkManager.Singleton.NetworkTimeSystem.ClientBufferTime` and `NetworkManager.Singleton.NetworkTimeSystem.ServerBufferTime` can be used to adjust the time. Note these settings should only be adjusted on a client. The server does not apply any buffering.
+Buffer time can be manually changed in code to for instance use different buffer sizes for different platforms. `NetworkManager.Singleton.NetworkTimeSystem.ClientBufferTime` and `NetworkManager.Singleton.NetworkTimeSystem.ServerBufferTime` can be used to adjust the time. Note these settings should only be adjusted on a client and must be adjusted on each client individually. The server does not apply any buffering.
 
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
@@ -143,9 +119,10 @@ The implementation of time proposed in this RFC is build around a star topology.
 
 This RFC introduces 3 new components to MLAPI. 
 
-- A **NetworkTime** struct. While time is just a double often when dealing with time a broader context is needed (access to last fixed Time, delta time, ticks etc.). The **NetworkTime** struct can be used to created that context from a double time value.
-- A **INetworkTimeProvider** will be added to MLAPI. This interface defines the implementation of a `NetworkTimeProvider`. The `NetworkTimeProvider` controls how the flow of time advances in the game and can adjust time depending on network conditions or other factors.
+
+- A **NetworkTime** struct. This struct allows us to expose time in a more familiar manner to the user in the form of a context which gives access to the last fixed Time, delta time, ticks etc.
 - A **NetworkTimeSystem** class will be added to MLAPI. The NetworkTimeSystem keeps track of networked times and ticks. Other systems can use the NetworkTimeSystem to access time information.
+- A **NetworkTickSystem** class will be added to MLAPI. The tick system takes time values from a time system and whenever the time passes a tick interval, it runs a tick.
 
 ## Network Time Struct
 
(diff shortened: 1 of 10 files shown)


text/0014-network-time.md
```diff
@@ -0,0 +1,683 @@
...
+    }
+
+    public void InitializeClient(ref NetworkTime predictedTime, ref NetworkTime serverTime)
+    {
+        predictedTime = serverTime + 0.1f;
+    }
+}
+```
+
+This `INetworkTimeProvider` is a very simple provider which increments `ServerTick` and `PredictedTick` at a consistent rate while running the `PredictedTick` 100 ms ahead of the `ServerTick`. This just acts as an example implementation and is not a recommended way to implement a `INetworkTimeProvider`
```
SamuelBellomo, line 129: "100 ms ahead of the ServerTick"
tick's unit shouldn't be time, it should be an int counting the number of network updates

LukeStampfli: Correct, ticks are ints and not floats. I can see why the explanation here is confusing. ticks are derived from time. Similar to how Unity decides on how many FixedUpdate states it runs in front of each update based on Time.time we do the same thing but with `PredictedTime`.

text/0014-network-time.md
```diff
@@ -0,0 +1,683 @@
...
+```
+or by accessing the `PredictedTime` and `ServerTime` properties of the `NetworkTimeSystem` directly.
+
+## Architecture
+
+![architecture](0014-network-time/architecture.png)
+
+- **INetworkTimeProvider** Open interface, in control of how time advances relative to Time.deltaTime.
+- **NetworkTimeSystem** Closed system to keep track of time, convert them to ticks. Also providers helper functions and APIs to interact with network time.
+- **NetworkManager** Executes logic for each network tick. Advances the `NetworkTimeSystem`.
```
MFatihMAR, line 352: If I understand this correctly, we're not implementing a new `INetworkUpdateSystem` and instead, `NetworkManager` will tick our `NetworkTickSystem` directly (in other terms, `NetworkUpdateSystem` is not `NetworkUpdateSystem : INetworkUpdateSystem`) right?

LukeStampfli: Correct, the reason why I did it like that is because it made sense to me to have `NetworkTimeSystem` decoupled from the player loop for multiple reasons
- we might want to reuse the time system in a separate scope. Replays, rollback maybe.. haven't thought much about it but I think there could be use cases for creating a separate time system and ticking it.
- gives us more control in the NetworkManager to run code before and after we advance the time system
- easier testing

I'm not very opinionated about this though. If we think that `NetworkTimeSystem` should be a `INetworkUpdateSystem` directly we can do that.

MFatihMAR: good to know. this is more or less an implementation detail — I was just curious, not have an opinion yet.

text/0014-network-time.md
```diff
@@ -0,0 +1,681 @@
...
+
+After applying the changes, it will look like this:
+
+<p float="left">
+<img src="0014-network-time/platform_3.1.gif" width="45%">
+<img src="0014-network-time/platform_3.2.gif" width="45%">
+</p>
+
+***Client** RTT = 200ms, `Own Player` = aligned, `Server NPC` = 200ms behind, `Other Player` = 200ms behind*<br>
+***Host** RTT = 200ms, `Own Player` = aligned, `Server NPC` = aligned, `Other Player` = aligned*
```
MFatihMAR, line 236: a minor markdown syntax fix:

```suggestion
_**Client** RTT = 200ms, `Own Player` = aligned, `Server NPC` = 200ms behind, `Other Player` = 200ms behind_\
_**Host** RTT = 200ms, `Own Player` = aligned, `Server NPC` = aligned, `Other Player` = aligned_
```

.gitattributes
```diff
@@ -0,0 +1,2 @@
+*.gif filter=lfs diff=lfs merge=lfs -text
+*.png filter=lfs diff=lfs merge=lfs -text
```
MFatihMAR, line 2: I don't like seeing GIt LFS around :/
I think we don't really need Git LFS.

text/0014-network-time.md
```diff
@@ -0,0 +1,681 @@
...
+### Predicted Time
+When moving your own player character locally, the character is ahead of time compared to all other NetworkObjects. This is the case because any changes to state will first need to be sent to the server and then back to all clients. `PredictedTime.Time` can be used to access this time. Most of the time `PredictedTime` should be used instead of `ServerTime` when implementing gameplay code.
+
+Example of using Predicted Time:
+
+```csharp
+public class NetworkChat : NetworkBehaviour
+{
+    [ServerRpc(RequireOwnership = false)]
+    SendChatMessageServerRpc(string message, float time)
```
Jura-Z, line 77: I ask you to change this and other absolute `time` to at least `double`s. You should never use `float` for absolute time.

For instance, if the app is storing time in float in seconds since app start time. 
The error in precision will be .976 msec after ~3 hours, and 7.81 msec after ~24 hours.

Source:
1. https://randomascii.wordpress.com/2012/02/13/dont-store-that-in-a-float/ 
2. https://twitter.com/ID_AA_Carmack/status/418158611664097280

Jura-Z (changes requested):

text/0014-network-time.md
LukeStampfli: Agreed we should use double. Internally we are using a tick + float offset representation so we never have large absolute float values but I think it could still make sense to fully move do double. Since we have the chance to start fresh here with the API what would you recommend?
1. Make everything exposed a double. Both absolute and delta values.
2. Absolute values as doubles, delta values as floats.
3. Similar to the Unity Time API provide default as float but additional `AsDouble` API.

Jura-Z: 3. is actually 2. Unity stores time as doubles only internally. It is just an API that truncates it to float (Time.time), and AsDouble gives it as-is.

Variant 2 looks right to me, but I don't have a strong opinion.

LukeStampfli:
> So I would like to talk about this API being a little more complicated then it should be. I am not sure I agree or see the value of having tick rate and tickes linked to NetworkTime as they are often two very different constructs which of course one can use NetworkTime to be the foundation of.
> 
> Second I am not sure I get why this API has to be as intertwined to the NetworkLoop or advancing time or being an interface or dealing with these "What if" use cases that we don't actually have a callout for. So I would like to purpose a different API that I have used for several real-time and turned based games.
> 
> 
> This API doesn't conflate Tick or Prediction etc... its very simple, it also doesn't rely on Unity Time.Time which is something I don't think is required. Also I think that we don't need an interface or advancing time in any special location etc... pretty basic stuff and this works for 90% of most real-time games and even non-realtime.
> 


Very good comment, gonna split this up a bit.

## Tick rate and ticks linked to NetworkTime
I do agree that ticks might should not be linked to NetworkTime. What we could do is have a NetworkTimeSystem and a NetworkTickSystem where we first advance time and then the tick system takes that as an input to execute the tick logic. I can definitely so why that would be useful. For instance if we do reconciliation we want to rerun ticks but not really touch our network time. So what do you think about the following architecture?
- NetworkTimeSystem: Runs network time by using absolute double values of time without tick involved.
- NetworkTickSystem: Uses NetworkTimeSystem or other inputs to run fixed ticks.
- NetworkTime: The API we use to easily expose time to the user so they have access to both time and fixed time. NetworkTickSystem will provide this value.

The reason I want NetworkTimeSystem to have access to ticks is because it is still driven by ticks. The values which drive the time system are RTT and ack times. We can decide to completely ignore this and run network time based on absolute floating time values. Or we could give the NetworkTimeSystem access to this information, how much time passed since the last tick? When could I theoretically expect the next tick etc. Right now we are not using that information in any INetworkTimeProvider which is a very valid reason to say we want to scrap this idea. The reason why I added this was because I have seen games use that information to drive time and I wanted to keep our options open.

## Why does the `NetworkTime` struct use ticks?

This struct is more of a `NetworkTimeContext` actually. It is an user facing API which tries to give a user all the functionality of Unity's Time API (access to Time, FixedTime, FixedDeltaTime). Why do we need this? MLAPI gameplay code might be run in different time spaces, for instance when we do client side prediction and reconciliation during the reconciliation our player character will run its update function in a different reconciliation time space. The `NetworkTime` struct allows us to easily pass that information as a parameter into a function.

## Why the INetworkTimeProvider Interface?
So the reason why we need this, I think is validated by this comment here. I have seen a variety of different approaches for doing network time. Your approach might be able to cover 90% of the use cases my approach might be able to cover another 90%. I do not think where we can get to a point where we have something which works with everyone.

With MLAPI we want to cover small scale co-op to hyper competitive games which want to minimize latency. We currently do not have a good time system for these kinds of games because they often drive the advancements of time based on the command buffer size on the server (we cannot do that yet as we do not have a command buffer). So in the future we might introduce a time provider which solves that issue but without the interface we would have to introduce all changes into the time system itself which could result into it growing into something which gets really hard to maintain.

If we decide to use a separate tick and time system then the argument for not needing INetworkTimeProvider gets much stronger.  We could just have a INetworkTimeSystem interface instead which would just declare `LocalTime` and `ServerTime` properties. The reason why I think we still need `INetworkTimeProvider` is because I think we need control over when network time changes and always want to do this at the beginning of the player loop. Basically we want to do the following in one frame:

1. Poll the transport > Update receive buffers, snapshots etc.
2. Advance the NetworkTime/ NetworkTicks
3. Run fixed network ticks
3. Run the Unity `Update`
4. Flush the transport

During 2-4 we do not want network time or ticks to change. So the idea is that our `NetworkTimeSystem` class ensures that the time is updated at the right time. `INetworkTimeProvider` is the interface which decides how time is updated. It could be using your implementation, it could be using mine, anything really.
 
 > This API doesn't conflate Tick or Prediction etc... its very simple
 
Just to clarify, my API does not have anything to do with Prediction. I just named your 'LocalTime' as 'PredictedTime' mainly because I wanted to follow DOTS Netcode naming conventions. But based on feedback so far we might change this to `LocalTime`

> it also doesn't rely on Unity Time.Time which is something I don't think is required

We also do not rely on unity Time.Time. The NetworkTimeSystem can be advanced with any time value. Right now we are using Time.deltaTime but it could be anything. The INetworkTimeProvider can always ignore the incoming delta time and enforce it's own time. So you could use your SystemTime based provider with the existing time system.

Why are we using Time.deltaTime as a default? From my very limited understanding of how time works inside the Unity engine I thought that using time values from an absolute timer like you are suggesting or by using Time.realtimeSinceStartup we will get inaccurate results. Reason being we do not care about how much time has actually passed in real life, we care more about. We care about 'Wall Time' or how much delta has passed between two render frames. Based on my understanding of [this article](https://blog.unity.com/technology/fixing-time-deltatime-in-unity-2020-2-for-smoother-gameplay-what-did-it-take) this is crucial for achieving smooth interpolation. Could be very well that I'm wrong about that.

LukeStampfli:
Some architecture changes have been made to the original RFC. NetworkTimeSystem will be a full standalone component using the NetworkUpdateLoop system and no longer dependent on NetworkManager advancing the time.

MFatihMAR:
@LukeStampfli 
I'd be curious if we could get a minimal and more focused NetworkTime proposal that targets our needs & features today and solves only those issues in the near term — then hopefully listing out, outlining our future plans and current thoughts but not trying to address them today (in other terms, not addressing future concerns from today but still thinking and noting things here).
Do you think that'd make sense or do you think we can't reduce our surface-level / scope of this RFC proposal?

LukeStampfli:
@MFatihMAR I cannot think of anything which could be removed to reduced the surface-level of this RFC. All the future plans mentioned will be implemented before our next release. The feature of the time system are needed to unblock work on my other tasks. I think the network time system could be restructured based on some of feedback from @wackoisgod. I agree with a lot of the points made but not sure whether I have the time to redesign this.

LukeStampfli:
I have moved this RFC into FCP with the goal of closing it to unblock future work. There are open comments in this RFC which could be addressed to improve this system:
- Splitting time and tick into two separate systems

- Simplifying the NetworkTimeSystem by removing the INetworkTimeProvider and maybe overall simplifying the implementation
I do not think there is an immediate need to address those. While splitting Network time and tick systems might lead to better separation of concerns it is always a change we can later as both are currently encapsulated in the NetworkTimeSystem. Whether the INetworkTimeProvider is needed might be questionable but I'd rather keep an open interface instead of restricting the time system and potentially releasing a time system which will not support certain archetypes.


In addition there are some UX questions open which I plan to address with a UX designer.

Areas which were addressed changed by this RFC discussion:
- NetworkTime is now internally stored as a double and we calculate everything using doubles. Some user facing APIs will expose floats, we will just cast the doubles for that.
- NetworkTimeSystem will be decoupled from NetworkManager. It will be able to run in any NetworkUpdateStage as a separate system.

Jura-Z (approved):

text/0014-network-time.md
```diff
@@ -0,0 +1,681 @@
...
+
+```csharp
+public static NetworkTime operator +(NetworkTime a, NetworkTime b)
+{
+    Assert.AreEqual(a.TickRate, b.TickRate, $"NetworkTimes must have same TickRate to add.");
+
+    int tick = a.Tick + b.Tick;
+    float tickDuration = a.m_TickDuration + b.m_TickDuration;
+
+    if (tickDuration > a.m_TickInterval)
```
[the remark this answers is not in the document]
LukeStampfli: I think you reviewed an old document here by accident. In the newest revision this complicated code is no longer necessary and I just do a.Time + b.Time to add two times together.

andrews-unity (approved):
I approve the new Time design!

PR #15, opened 2021-05-24
Authority

LukeStampfli:
[See linked RFC document →](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/authority/0015-authority.md)

RFC proposing an authority system.


0015-authority.md
```diff
@@ -0,0 +1,478 @@
...
+            }
+        }
+    }
+}
+```
+*The authority system allows NetworkVariables to be flexibly used as inputs.*
+
+## Changing Authority
+
+The `Authority` mode of a `NetworkObject` can be changed, on the server only, by calling `networkObject.ChangeAuthority(Authority.Server);`.
```
SamuelBellomo, line 125: mm that part I don't agree. To me, a client authoritative cube should only be changeable by the client that owns it, not the server. Else you could easily run into issues where a client still thinks it has authority over the cube and take decisions over it while the server has already sent your a delayed "authority change" message. Since the client is authoritative, it shouldn't have correction mechanisms to correct its own authoritative changes on that cube.

0015-authority.md
```diff
@@ -0,0 +1,478 @@
...
+- Owner: The owner of the `NetworkObject` has authority over it.
+- Server: The server always has authority over the `NetworkObject.
+
+![dropdown NetworkObject](./0015-authority/dropdown-ownership-networkobject.png)
+
+We might add new authority modes in the future.
+
+Having Authority over a `NetworkObject` does the following
+- There is a `HasAuthority` property on the `NetworkObject` and each `NetworkBehaviour` which will be set accordingly.
+- Built in MLAPI components such as `NetworkTransform` or `NetworkAnimator` will synchronize the state of the at
```
SamuelBellomo, line 51: of the at?

0015-authority.md
```diff
@@ -0,0 +1,478 @@
...
+- NetworkVariables with `ReadPermission = NetworkVariablePermission.Authority` can be read from. (write permission also exists)
+
+Having Authority over a `NetworkObject` **does not**: change MLAPI features such as how NetworkVariables or RPCs are sent/received etc. In the future we might have authority modes such as `Predicted` which will change this. But this is not part of the authority mode but will be implemented by the specific feature like `prediction`.
```
SamuelBellomo, line 54: "predicted" would still remain server authoritative in that case. I'm not sure why the authority switch would be aware of that. Do you mean that a server authoritative netvar would prevent updating that netvar client side? I would keep it like we have right now, where the server throws errors when receiving updates from netvars it shouldn't.
This way client side you can still allow changing a server authoritative netvar, but don't sent the value change if you're predicted.

0015-authority.md
```diff
@@ -0,0 +1,478 @@
...
+    Debug.Log("Authority Lost :(");
+}
+```
+
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## Why is Authority a mode?
+Why is authority not something like ownership which you give to a client. The reason for that is because authority is a relation between multiple peers. In an authoritative model a set of peers pass inputs to the authoritative peer for processing.
```
SamuelBellomo, line 149: "In an authoritative model a set of peers pass inputs to the authoritative peer for processing."
not with client authority. then each clients would use their own inputs and only send their authoritative state to the server.

0015-authority.md
```diff
@@ -0,0 +1,478 @@
...
+This RFC will introduce some rules and responsibilities for each of the two systems.
+
+IsOwner:
+- Only one peer can own a NetworkObject.
+- The `owner` of a NetworkObject can use `ServerRpcs` on its `NetworkBehaviour` by default.
+- The `owner` is responsible for collecting inputs and redirecting them to the authority.
+- The `owner` is encouraged write to `NetworkVariables` with readpermission = authority but not to other NetworkVariables.
+- When the owner peer disconnects, the object it owns might get destroyed when `DontDestroyWithOwner` is set to false.
+
+HasAuthority:
```
SamuelBellomo, line 167: I'd be interested to put into writing the relation between IsOwner and HasAuthority here

0015-authority.md
```diff
@@ -0,0 +1,478 @@
...
+Accepting inputs from multiple peers is something we will not support with the current proposed design. The authority mode system allows for further extension to support those modes.
+
+## Ownership vs Authority
+This RFC will introduce some rules and responsibilities for each of the two systems.
+
+IsOwner:
+- Only one peer can own a NetworkObject.
+- The `owner` of a NetworkObject can use `ServerRpcs` on its `NetworkBehaviour` by default.
+- The `owner` is responsible for collecting inputs and redirecting them to the authority.
+- The `owner` is encouraged write to `NetworkVariables` with readpermission = authority but not to other NetworkVariables.
```
SamuelBellomo, line 164: you mean "writepermission = authority", not read permission?

0015-authority.md
```diff
@@ -0,0 +1,478 @@
...
+- When the owner peer disconnects, the object it owns might get destroyed when `DontDestroyWithOwner` is set to false.
+
+HasAuthority:
+- The `authority` should run gameplay code such as physics, movement, path finding, animation etc.
+- The `authority` should update NetworkVariables which get used by other clients to visualize the game state.
+
+IsServer:
+- The server has permission to change ownership and authority of NetworkObjects.
+- `IsServer` should be avoided for gameplay related code.
```
SamuelBellomo, line 173:
```suggestion
- `IsServer` should be avoided for gameplay related code. HasAuthority should be used instead.
```

0015-authority.md
```diff
@@ -0,0 +1,478 @@
...
+        throw new SpawnStateException($"Cannot get authority for not spawned {nameof(NetworkObject)}");
+    }
+}
+```
+
+There will also be a function to change authority:
+```csharp
+public void ChangeAuthority(Authority authority)
+{
+    Assert.IsFalse(authority == Authority.FromNetworkObject)
```
SamuelBellomo, line 198: If I want to put it back to FromNetworkObject after changing it to something else, why not allow it?

0015-authority.md
```diff
@@ -0,0 +1,478 @@
...
+This system is very limited and not really `authority` it is more what we have in MLAPI with `ownership`. It looks like the MLAPI `ownership` model has been inspired by the UNet `authority` model. The MLAPI `authority` model will improve MLAPI and allow users to do things which were not possible in UNet.
+
+## DOTS Netcode
+DOTS Netcode has a ghost mode which can be interpolated or predicted. The interpolated mode matches our 'server' mode. There is no mode for client authority. The predicted mode is something we do not support yet but can in the future.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+## Scrapping per NetworkBehaviour Authority
+See appendix for details. There are very little actual use cases for this and scrapping this will more then half the code size of the authority feature and lower the bandwidth usage of MLAPI.
```
SamuelBellomo, line 297: Is this something Gio could help with, finding whether there are use cases for this?

0015-authority.md
```diff
@@ -0,0 +1,478 @@
...
+
+## Snapshots / AOI
+There are a lot of unknowns on how our snapshot and AOI system will work with `NetworkVariablePermission.Authority`. We do not plan to resolve this unresolved question in the authority RFC as it only introduces an additional `NetworkVariablePermission`. The snapshot / AOI RFCs will answer this question.
+
+## AuthorityChanged Event
+Do we need an AuthorityChanged event? We already have `OnAuthorityGained` and `OnAuthorityLost` events but would we need an additional to event to signal if the authority mode itself changed?
+
+## Default NetworkVariable Permissions
+
+The default write permission for a NetworkVariable is currently `NetworkVariablePermission.ServerOnly` and the default write permission is `NetworkVariablePermission.Everyone`. We could consider changing the default write permission to `NetworkVariablePermission.Authority`. It is not clear whether that would simplify or complicate the onboarding experience of MLAPI.
```
SamuelBellomo, line 310: If we change the default to authority, we would need to have the default authority on NetworkObject be server. This wouldn't **actually** change the default behaviour

0015-authority.md
```diff
@@ -0,0 +1,478 @@
...
+- Start Date: 2021-05-19
+- RFC PR: [Unity-Technologies/com.unity.multiplayer.rfcs#0015](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/pull/0015)
+- Issue: [Unity-Technologies/com.unity.multiplayer#0000](https://github.com/Unity-Technologies/com.unity.multiplayer/issues/0000)
+
+# Summary
+[summary]: #summary
+
+The authority concept is an abstraction layer for gameplay logic on top of MLAPI. It allows high level components to write game play code without the need of considering the underlying network model. Authority does not directly affect other parts of the framework.
+
+# Motivation
```
MFatihMAR, line 11: I can see this proposal from a technical perspective but I'm struggling to understand what are the real motivations behind this. Instead of "here's a concept that might be useful", I'd like to see something like "here's a concept that _is_ useful because we will use it to solve X problem in Y context".
I got "what?" and "how?" but I couldn't get "why?" and "when?".

0015-authority.md
```diff
@@ -0,0 +1,478 @@
...
+{
+    SetPosition(m_TargetPosition);
+}
+else
+{
+    SetPositionServerRpc(m_TargetPosition);
+}
+```
+
+In the future we could consider introducing an `AuthorityRpc` or similar which sends an RPC to the authority.
```
SamuelBellomo, line 339: Something interesting with this is you wouldn't have to care about whether you're sending an RPC to a server or a client. If I have a client authority, that RPC would use the current object's owner to know to which client it should send the RPC.
I feel like having that concept of authority wouldn't be complete without RPCs. You cover netvars only here, but from a user's point of view, I would expect Authority to work with both Netvars and RPCs, no?

0015-authority.md
```diff
@@ -0,0 +1,478 @@
...
+
+The RFC does not cover authority modes where multiple peers have authority over an object. It also does not cover prediction. The proposed authority system can be extended to support more authority modes. To support shared/distributed authority we might need an ACL entry to indicate which peers have authority over an object
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+## Prediction
+The authority feature is a first step towards the goal of supporting client side prediction. When using client side prediction we need precises control over what's being run on the predicted client and run the same on the server. The `HasAuthority` flag will allow us to do this by setting it to true on both the predicting client and the server.
+
+## Commands
```
MFatihMAR, line 322: I can't really see too much difference between commands vs RPCs. It feels like a more dedicated & specialized RPC implementation which brings me to the fact that we could simply extend and leverage existing RPC infrastructure.
Also, another note, both `NetVar` and `RPC` are just transport utilities, there is nothing wrong with carrying input data over an RPC and changing/simulating state on the server-side, then replicating server-sealed state back to all clients. There are no technical limits implementing input/predict/simulate/correct mechanism in the combination of `ServerRPC`s + pos/rot/... `NetVar`s.

0015-authority.md
```diff
@@ -0,0 +1,478 @@
...
+```csharp
+public Authority Authority => m_Authority;
+
+public bool HasAuthority
+{
+    get
+    {
+        if (IsSpawned) // This is needed because else we don't know whether we are a server.
+        {
+            return IsOwner ? Authority == Authority.Owner : NetworkManager.IsServer && Authority == Authority.Server;
```
MFatihMAR, line 187: This is obviously a form of syntactic sugar and I question do we really need this or should we just put this syntactic sugar into our specific implementations, as per one of your examples, `NetworkTransform`. Simply, `NetworkTransform` could have this getter-only property (of `HasAuthority()` method) on its own and we wouldn't touch anything at a lower-level-layers and wouldn't expose some extra public API we would maintain for a while.

0015-authority.md
```diff
@@ -0,0 +1,478 @@
...
+
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- Why is this design the best in the space of possible designs?
+- What other designs have been considered and what is the rationale for not choosing them?
+- What is the impact of not doing this?
+
+Authority could have been done as a permission/acl system instead of an authority mode. This would have allowed us to have more fine-grained control over who has authority. We were not able to find many real world use cases where this would be useful so that's why the authority mode was chosen. In the future we might consider storing authority internally as a set of owners instead of an authority mode but then most likely we would still want to expose authority as an authority mode for ease of use.
```
MFatihMAR, line 277: Well, I think ACL wouldn't even need to be exposed to the end-user directly. It could be a backbone that we use internally and extend over time with more props, flags, queries etc.
In the absence of a full-fledged and generalized ACL concept within MLAPI internals, I believe we could address these usage-patterns with ad-hoc implementations as @SamuelBellomo did in `NetworkTransform`.
The reason why I'm saying that is basically the fact that you also mentioned this Authority Model wouldn't fit in all places and we'll probably keep Ownership-over-RPC and most parts of NetVar-ReadWrite-Perms for a while.
I'm questioning whether we add more things to replace later here, or in other terms, I can't really see this model being durable for a long tme.

MFatihMAR:
In general, you can see my suspicious comments about the proposed Authority model, its coverage, immediate use-case, durability etc.
I might be missing a bigger context here but nevertheless, I'd like to see more concrete materials in those sections.
I'm not saying "no", I'm saying "maybe" and I want to say "yes", hopefully :)

0015-authority.md
LukeStampfli: I cannot implement NetworkPhysics  without this. We cannot  provide NetworkTransform or NetworkAnimator without this. This is solving real problems we have right now.

0015-authority.md
LukeStampfli: The motivation section explains this. If we do not implement this `NetworkTransform` cannot work with other components such a `NetworkAnimator`. `Authority` is a way for a NetworkBehaviour to communicate how it works and what network model it supports. If  each component needs to implement its own form of this concept we end up with a lot of duplicate code and lose many of the upsides mentioned in #motivation

0015-authority.md
LukeStampfli: Command is more of a concept and not a specific implementation.
- RPCs can be commands
- NetworkVariable (state) can be commands.

Both have advantages/disadvantages in some situations. State can be rolled back, delta compressed and stored in the snapshot system. RPCs are much more flexible for one time commands. Most games get much better result by using state based commands for regular gameplay. DOTS Netcode also supports both ideas: <https://docs.unity3d.com/Packages/com.unity.netcode@0.0/manual/command-stream.html>

There are no hard technical limits which prevent RPCs commands to be predicted but we would need to introduce significant changes to RPCs. NetworkVariables do not need those changes.

0015-authority.md
LukeStampfli: Oh you are right, this is just a stupid pre-condition, ignore it 😄

0015-authority.md
LukeStampfli: In my mind with client authority the owner peer passes inputs to itself (the authoritative peer). But yeah I can see why the wording might be confusing.

0015-authority.md
LukeStampfli: Oh I did not think about that. I kind of started with the assumption that the default authority is `owner` but not because I had a good reason. After thinking about it having `server` as the default authority makes sense to me.

0015-authority.md
MFatihMAR: Do you mind mentioning and explaining those real problems within the RFC?

0015-authority.md
MFatihMAR: Where in the motivation section do you touch on these details?

0015-authority.md
LukeStampfli: > This way client side you can still allow changing a server authoritative netvar, but don't sent the value change if you're predicted.

This is was what I was trying to say. If we do prediction one way of doing this would be as you mentioned to allow the client to write but then not synchronize those changes.

0015-authority.md
LukeStampfli: I mean `readpermission = authority` as it is there. The `owner` can use NetworkVariables with `readpermission = authority` to send inputs to the authority. That was what I was trying to say here. For this to work the writepermission would have to be set to `owner` of course.

0015-authority.md
LukeStampfli: @UnityGio for visibility

0015-authority.md
LukeStampfli: So I feel like no matter which default we chose we will break some stuff:

- if we set owner as default => NetworkVariables will behave differently when we set the default permission to `NetworkVariablePermission.Authority`
- if we set server as default => Components like NetworkTransform which are currently owner authorative by default will now be server authoritative by default.

To me this is not a problem and we should rather clean this up now than later. Just pointing it out. For me 'server' makes more sense as the default authority value but feels like a UX question at this point?

LukeStampfli:
I moved the `SupportedAuthorityAttribute` from the appendix back into the main RFC. This attribute is useful even if we do not have per NetworkBehaviour authority. It was moved into the appendix by mistake.

0015-authority.md
SamuelBellomo: you're talking about "read" and "sending inputs". To me those two are in opposition. reading shouldn't imply any changes.

0015-authority.md
LukeStampfli: I want to point out that this issue was already a problem with Ownership before we introduced Authority. If the owner of a NetworkObject was changed by the server the client who owned it still had `IsOwner` true and was able to write to NetworkVariable with ReadPermission = Owner.

I believe we want the server to always be able to take ownership or authority away because else we allow potential exploits/ hacks. So we will need a solution for this case. I cannot think of a solution which we can implement right now. I think the proper way to solve this is to do a state rollback.
1. Server changes authority of an object and sends this change to everyone. This change is associated with a specific tick
2. If a client receives this change it will check if it lost authority over an object. If true it will roll back the state of the NetworkObject to that given tick where the authority changed and clean the snapshot of all "wrong" future values.
3. The server will ignore incoming NetworkVariable changes sent in this transition period by the "wrong" authority.

0015-authority.md
LukeStampfli: Let me try to reformulate this. Does this make more sense?
> If the `owner` wants to pass inputs to the authority this can be done by writing to `NetworkVariables` with WritePermission = owner and ReadPermission = Authority.

0015-authority.md
LukeStampfli: Added a short paragraph and my explanation to the RFC document

0015-authority.md
LukeStampfli: Only in my head apparently 😅. I'll rework the motivation section.

0015-authority.md
LukeStampfli: I will add more information to the motivation section to explain this.

0015-authority.md
SamuelBellomo: You're assuming the server _should_ change the authority. in a pure client authoritative model, only the client could change it, the server would only be syncing these changes to other clients and could do any modifications to it.
else, you're making client authoritative games have to handle rollback, making it not client authoritative anymore

0015-authority.md
LukeStampfli: fixed

0015-authority.md
```diff
@@ -0,0 +1,501 @@
...
+            m_AuthorityModes = authorityModes;
+        }
+    }
+```
+
+At runtime modes get cached in a static lookup table `Dictionary<Type, Authority>` to avoid reflection in the hot path.
+
+Supported Authority mode just changes some minor details in NetworkObjectEditor to display components with not supported authority and after changing authority it will check whether the behaviour is still supported. In addition it checks foreach NetworkBehaviour whether the authority mode is still allowed when authority gets changed and if not logs a warning to the console.
+
+# Drawbacks
```
ShadauxCat, line 308: I think it's worth calling out in the drawbacks that inexperienced developers might be confused by the difference between ownership and authority. I don't think that's a reason not to implement authority, but it is definitely a drawback to making that separation and will need to be well-documented.

0015-authority.md
0x-1: My guess is that it would end up in something like this:

Authority/Control gives a client possibility for input/output and temp. state (aka local write trough)
Owners have the final say about a state of an Object disregarding but maybe including Authority.

Scenarios for abstraction:
In context of movement this can give a client predictive abilities as it can set input/output and temp. state (client prediction) without being the owner(server-auth)
In context of an inventory this can make input-lag-less UI possible (anolog to client prediction) without being the owner(server-auth)

In those scenario cases either client or server could be the owner but it would not change the context.
Thus a correct abstraction away from isClient / isServer.

conclusion
A)  Maybe we need a more consistent naming?
-> Control vs Owner
-> Input Auth vs State Auth
-> Simulant vs Original (thus opening replica/copy nameing on the other maschines)

B) Networking Mode specific defaults?
-> Dedicated Servers would have Ownership over all objects but give Auth to specific clients.
-> Client/Host-Mode: Each client has both Ownership and Auth over specific Objects while Server is only messenger.

C) This would end up requiring either an Acknowledge or a state-rollback(like you commented already) if the owner decides against a change. Maybe we end up in something like Acknowledge First Authority VS Roll-Back Authority.
This RFC decided for a simple Authority check but misses out on the roll-back part (thus still requires rpcs)

LukeStampfli:
Closing this for now as needs more discussion

PR #16, opened 2021-05-25
NetworkSpawn

LukeStampfli:
[See linked RFC document →](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/network-spawn/text/0016-network-spawn.md)

RFC proposing renaming `NetworkStart` to `OnNetworkSpawn` and introducing a `OnNetworkDespawn` virtual event function.

### DIFF
--- a/text/0016-network-spawn.md
+++ b/text/0016-network-spawn.md
@@ -1,0 +1,76 @@
+- Feature Name: OnNetworkSpawn
+- Start Date: 2021-05-25
+- RFC PR: [Unity-Technologies/com.unity.multiplayer.rfcs#0016](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/pull/0016)
+- Issue: [Unity-Technologies/com.unity.multiplayer#0000](https://github.com/Unity-Technologies/com.unity.multiplayer/issues/0000)
+
+# Summary
+[summary]: #summary
+
+Rename `NetworkStart` to `OnNetworkSpawn` and introduce `OnNetworkDespawn` function.
+
+# Motivation
+[motivation]: #motivation
+
+- Many users try to subscribe to events in `NetworkStart` but then have no place to unsubscribe. As a user I want to know when my NetworkObject gets spawned and despawned. The proposed API will make this accessible.
+- `NetworkStart` might lead people to think that it is a similar initialization function as `Start` or `OnEnable` which is not true. The new wording will be more direct.
+- This will also make pooling/addressable support etc. much more straight forward because we decouple object life cycle from network lifecycle.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+`NetworkStart` has been renamed to `OnNetworkSpawn`. `OnNetworkSpawn` gets called when a server spawns an object on the server and gets called when a client replicates the spawn of a NetworkObject.
+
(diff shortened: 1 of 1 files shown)


text/0016-network-spawn.md
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    private void MyValueChangeFunction(float prev, float next)
+    {
+        Debug.Log($"previous {prev}, next {next}");
+    }
+}
+```
+
+# Reference-level explanation
```
MFatihMAR, line 48: makes a lot of sense, thank you for this RFC.
I only have one question, do you mind outlining where/how you'd wire `OnNetworkDespawn` events on the server and clients?
Some neighbor codes and/or call-stack to get to the `OnNetworkDespawn` would be great!

LukeStampfli: I added this to the reference level section of the RFC. I also plan to do some small refactoring around how NetworkStart is called (no behaviour changes just throwing out unused legacy code)

text/0016-network-spawn.md
```diff
@@ -0,0 +1,90 @@
...
+[reference-level-explanation]: #reference-level-explanation
+
+Both the internal `InternalNetworkStart` and virtual `NetworkStart` will be renamed to the new naming convention. In addition we will introduce the same two functions "`InternalNetworkDespawn`, `OnNetworkDespawn`" to NetworkBehaviour.
+
+The internal function will be used so that NetworkBehaviours can subscribe to internal MLAPI events.
+
+`OnNetworkDespawn` will also be called on the server when a `NetworkObject` gets destroyed because that also counts as a despawn.
```
TwoTenPvP, line 55: Surley this should only occur if IsSpawned is true? Otherwise you could have a NetworkObject that is never actually spawned, (thus OnNetworkSpawn is never called) but OnNetworkDespawn is called.

mariodebono: `OnNetworkDespawn` is very interesting to me, would this also give me room to gracefully fade away the object over some time on client or the object would be destroyed anyway before I could fade it?

LukeStampfli: You won't be able to use `OnNetworkDespawn` for this specific use case because MLAPI would still force destroy the object after. In this case you would need to implement a custom destroy handler to override the destroy logic of MLAPI:
<https://docs-multiplayer.unity3d.com/docs/advanced-topics/object-pooling>

@TwoTenPvP Yes, correct this will only occur if IsSpawned is true.

mariodebono: @LukeStampfli, Makes sense. thanks for pointing out the object-pooling. 

Would `OnNetworkDespawn` also be called when server runs Show/HideNetwork?

MFatihMAR (approved):
LGTM! 🚀

Though, only one side-question: Do you have any plans to let docs people know and help them get things up-to-date with this upcoming change?

text/0016-network-spawn.md
LukeStampfli: Currently `NetworkShow` and `NetworkHide` spawn/despawn an object on the client and would trigger the`
 `OnNetworkSpawn` and `OnNetworkDespawn` functions

JesseOlmer (approved):
lgtm

ShadauxCat (approved):

PR #17, opened 2021-05-31
NetworkObject Parenting

MFatihMAR:
[See the linked RFC document →](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/object-parenting/text/0000-networkobject-parenting.md)

This RFC proposal is an attempt to outline basic needs for a `NetworkObject` (re)parenting managed by the server at both scene authoring time and at runtime.
There are a couple of limitations, however, we still want to deliver this functionality within MLAPI which would work out-of-the-box for some use-cases and some non-expert network programmers.

### DIFF
--- a/text/0000-networkobject-parenting.md
+++ b/text/0000-networkobject-parenting.md
@@ -211,17 +211,12 @@ Both UNet and Unreal does not offer a similar solution out-of-the-box and mostly
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
 
-- What parts of the design do you expect to resolve through the RFC process before this gets merged?
-- What parts of the design do you expect to resolve through the implementation of this feature before stabilization?
-- What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
+- Why can't we queue up all the changes to a buffer in non-networked gameplay and simply sync all of them on connect/join?
+    - Technical MLAPI limitations, see details above in guide & reference level explanation sections.
+- Why `NetworkObject` parenting can't be client-driven?
+    - Similar to "Ownership" model we have, we trust server to be the source of truth and we simply rely on server-side logic for both ownership and here for parenting too.
 
 # Future possibilities
 [future-possibilities]: #future-possibilities
 
-Think about what the natural extension and evolution of your proposal would be and how it would affect the Unity Multiplayer as a whole in a holistic way. Try to use this section as a tool to more fully consider all possible interactions with the Unity Multiplayer in your proposal. Also consider how the this all fits into the roadmap for the project and the team.
-
-This is also a good place to "dump ideas", if they are out of scope for the RFC you are writing but otherwise related.
-
-If you have tried and cannot think of any future possibilities, you may simply state that you cannot think of anything.
-
-Note that having something written down in the future-possibilities section is not a reason to accept the current or a future RFC; such notes should be in the section on motivation or rationale in this or subsequent RFCs. The section merely provides additional information.
+Potentially, we'll rework low-level core parts of MLAPI which would involve `NetworkObject`, `NetworkBehaviour` etc. and that'd also throw away all scene transform hierarchy related logic out and bump this user-need/feature to a higher-level construct like `NetworkTransform` in the future.
--- a/text/0000-networkobject-parenting.md
+++ b/text/0000-networkobject-parenting.md
@@ -1,7 +1,7 @@
 - Feature Name: `networkobject-parenting`
 - Start Date: 2021-05-26
-- RFC PR: [Unity-Technologies/com.unity.multiplayer.rfcs#0000](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/pull/0000)
-- Issue: [Unity-Technologies/com.unity.multiplayer#0000](https://github.com/Unity-Technologies/com.unity.multiplayer/issues/0000)
+- RFC PR: [RFC#17](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/pull/17)
+- Issue: [MLAPI#876](https://github.com/Unity-Technologies/com.unity.multiplayer.mlapi/issues/876)
 
 # Summary
 [summary]: #summary
--- a/text/0000-networkobject-parenting.md
+++ b/text/0000-networkobject-parenting.md
@@ -1,4 +1,4 @@
-- Feature Name: `object-parenting`
+- Feature Name: `networkobject-parenting`
 - Start Date: 2021-05-26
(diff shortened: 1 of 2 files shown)


text/0000-networkobject-parenting.md
```diff
@@ -0,0 +1,222 @@
...
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- Given the limited time and scope, we think we cannot re-architect MLAPI fundamentals to support an approach where `NetworkObject` and `NetworkBehaviour` components are simply unaware of scene and transform hierarchy but they are plus parenting to be provided by a higher-level construct outside MLAPI core such as `NetworkTransform`. We wanted to move forward with what we already have.
```
LukeStampfli, line 204: I'm just wondering. If we moved this feature into a separate high level NetworkBehaviour `NetworkParenting` and just told users to add this to every `NetworkObject` root if they want automagical parenting. How would that increase scope? I do not see any technical blockers to do that? Wouldn't that be a better approach since we cannot re-architect MLAPI fundamentals?

text/0000-networkobject-parenting.md
```diff
@@ -0,0 +1,222 @@
...
+
+### Only Reparenting Under A NetworkObject (Or To The Root) Is Valid
+
+A `NetworkObject` can only be reparented under another `NetworkObject` (`GameObject` with `NetworkObject` component attached). Only exception is moving a `NetworkObject` to the root of the scene hierarchy.
+
+This is simply due to the fact that MLAPI would not be able to identify & locate new parent on the remote-side if it was a non-`NetworkObject` parent. Again, except moving it to the root because we could identify no parent (root) scenario without `NetworkObject` identification or scene hierarchy traversal.
+
+### Only Reparenting During Networking Is Valid
+
+A `NetworkObject` can only be reparented while networking, in other terms you can only reparent while listening/running as a server.
```
LukeStampfli, line 43: Can an unspawned `NetworkObject` on the server be reparented while the server is listening/running?

MFatihMAR: nope.

text/0000-networkobject-parenting.md
MFatihMAR: we can't because if we were to move a `NetworkObject`, that info needs to be written down to the `NetworkObject` itself — it's the main identification/linking layer between local and remote `NetworkObject`s. Even if we were to move the actual logic implementation up to the `NetworkTransform` (or `NetworkParenting` level), we still need to sync `NetworkObject`s, not `NetworkTransform`/`NetworkParenting`. At login, we still need to serialize `NetworkObject`'s `m_IsReparented`, `m_LastestParent` etc.
having said that, all this feature design is subject to change when we start working on bigger refactorings and architectures. so, even though I'm not entirely happy with this proposal, I'm not too bothered to make it "just right" and/or "fancy" for now since all of it is temporary.

text/0000-networkobject-parenting.md
```diff
@@ -0,0 +1,222 @@
...
+# Motivation
+[motivation]: #motivation
+
+We want to offer `NetworkObject` reparenting solution within MLAPI to help developers synchronizing `transform` parent-child relationships of `NetworkObject`s.
+
+It is often not easy to grok the parenting in networking, also your choice of networking framework would have its limitations you don't necessarily know up until trying to implement and cover the edge cases. Here, we want to provide a solution that plays nicely with current MLAPI limitations and we expect this "NetworkObject Parenting" concept to evolve and change over time when MLAPI matures.
+
+With this RFC implemented, we expect developers to reparent their `NetworkObject`s at runtime during networked gameplay and synchronize across the network (server to clients) out-of-the-box within certain limits and conditions.
+
+# Guide-level explanation
```
MFatihMAR, line 22: I'm going to inline Luke's concern we discussed recently here and address it later.
We want to have an escape hatch to skip this feature entirely for expert netcode devs who know what they are doing. My immediate thinking is to offer that escape hatch via a flag on NetworkObject and let devs opt-out from auto-networkobject-parenting. I'll inline that solution and give a little more details soon.

MFatihMAR: done.

text/0000-networkobject-parenting.md
```diff
@@ -0,0 +1,226 @@
...
+[guide-level-explanation]: #guide-level-explanation
+
+## Opt-out
+
+Before we dig deeper into this feature proposal, it needs to be said that this feature is going to be behind a bool flag that can be toggled on the `NetworkObject` inspector UI, which will be enabled by default but experienced developer can opt-out from it and wouldn't be running any of this logic if they were to implement a solution on their own.
```
MFatihMAR, line 27: @LukeStampfli

text/0000-networkobject-parenting.md
```diff
@@ -0,0 +1,226 @@
...
+It is often not easy to grok the parenting in networking, also your choice of networking framework would have its limitations you don't necessarily know up until trying to implement and cover the edge cases. Here, we want to provide a solution that plays nicely with current MLAPI limitations and we expect this "NetworkObject Parenting" concept to evolve and change over time when MLAPI matures.
+
+With this RFC implemented, we expect developers to reparent their `NetworkObject`s at runtime during networked gameplay and synchronize across the network (server to clients) out-of-the-box within certain limits and conditions.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## Opt-out
+
+Before we dig deeper into this feature proposal, it needs to be said that this feature is going to be behind a bool flag that can be toggled on the `NetworkObject` inspector UI, which will be enabled by default but experienced developers can opt-out from it and wouldn't be running any of this logic therefore they could to implement a solution on their own as well.
```
MFatihMAR, line 27: @LukeStampfli

ShadauxCat, line 27: I wonder if it's better to make this opt-out or opt-in. I can see arguments for both. I'm not taking any firm position here, but feel like it's worth discussing the merits of each side.

On the one hand, making it opt-out means that reparenting will work correctly without any effort from the developer (so long as they use it correctly, which isn't necessarily safe to assume here). But, on the other hand, given the limitations of automatic reparenting and the number of situations it doesn't work, making it opt-out may effectively result in giving developers rope to hang themselves with.

With the limitations, it might be better to make it opt-in, because the very fact that it requires that opt-in would be a signal to developers that there are limitations and gotchas and should be an incentive to read the documentation before enabling it. Reading the documentation for this seems like it'll be critical to make it work correctly, so having that checkbox as a funnel toward documentation (maybe even with a link to the documentation page in the help tooltip?) could improve developers' rate of success in using this feature.

Just food for thought, so to speak.

LukeStampfli (approved):
RFC looks fine to me when keeping in mind that we will remove / rework this again later.

I think some of the special cases such as "orphans" etc. could be quickly described in this RFC but besides that it's all good.

PR #18, opened 2021-06-03
Message Ordering Guarantee

MFatihMAR:
[See the linked RFC document →](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/message-ordering-guarantee/text/0018-message-ordering-guarantee.md)

RFC proposing stronger ordering guarantees for reliable sequenced messages - specifically, guaranteeing MLAPI will not itself reorder any messages before delivering them to the kernel.


text/0000-message-ordering-guarantee.md
```diff
@@ -0,0 +1,171 @@
...
+{
+    ulong newOwnerClientId = GetNewOwnerClientId();
+    NetworkBehavior networkBehavior = m_obj.GetComponent<NetworkBehavior>();
+    networkBehavior.ChangeOwnership(newOwnerClientId);
+}
+```
+
+In this example, because `Spawn()` is called during `Update()` on the server, and `ChangeOwnership()` is called during `LateUpdate()` on the server, the same will be true on the client - the object will be spawned on the client during the `Update` stage, and its ownership changed during `LateUpdate`.
+
+# Reference-level explanation
```
MFatihMAR, line 90: I think there would be a wider and more fundamental refactor which would be to remove `InternalMessageSender` and `InternalMessageHandler` classes entirely. We stripped down both classes, especially `InternalMessageSender` is currently very slim. If we could also make these two types a part of this refactor work, that'd be super nice as well. I believe we could kill IMS easily, IMH maybe not so easily but having a look at that IMH class and its methods would be nice, at least to give us an idea about what it'd take to kill it.
Having said that, we'll ultimately formalize our network frame that starts with "network input" and ends with "network output" — during a frame everything would be just queued to our network output buffer. We're not there yet but I see this step being a positive step towards that direction, I like it!

ShadauxCat: I think it'll definitely be easy to remove IMS with these changes. I'll investigate removing IMH as well.

text/0018-message-ordering-guarantee.md
```diff
@@ -0,0 +1,171 @@
+- Feature Name: `message_ordering_guarantee`
+- Start Date: 2021-06-02
+- RFC PR: [RFC#18](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/pull/18)
+- Issue: [MLAPI#700](https://github.com/Unity-Technologies/com.unity.multiplayer.mlapi/issues/700)
+
+# Summary
+
+This feature adds a guarantee of ordering between all types of messages, that, unless the user requests otherwise, all messages sent over a **reliable sequenced** channel will be received and processed by the recipient process in the order they were sent by the sender process. It guarantees that MLAPI itself will not internally change the order of any messages, that they will all reach the transport level in the same order they were initiated/requested. This RFC does **not**, however, make any guarantees on ordering for messageds sent on non-sequenced channels, nor does it provide any new enforcement to force any messages to be sent on sequenced channels.
```
NoelStephensUnity, line 8: > on ordering for messageds sent
Minor typo should be "messages"

ShadauxCat: Thank you! Fixed.

text/0018-message-ordering-guarantee.md
```diff
@@ -0,0 +1,171 @@
...
+
+2. The second, smaller refactor is changing the default value of the UpdateStage value in the send parameters so that it reads from the current update stage rather than defaulting to 0 (Update).
+
+The vast majority of the work is in the former refactor, but the biggest user-facing change is the latter one.
+
+In order to complete this refactor, a few things need to happen:
+
+- All RPC-related classes will need to be renamed from Rpc\* to Message\*. This accounts for RpcFrameQueueItem, RpcQueueContainer, RpcQueueProcessor, RpcBatcher, and RpcQueueHistoryFrame.
+
+- All internal commands must be sent through `RpcQueueProcessor.SendFrameQueueItem()`. As a consequence, this also means all internal commands must now build `RpcFrameQueueItem` structs.
```
NoelStephensUnity, line 104: This might be a good place to add an "RpcFrameQueueItem" (perhaps call this MessageFrameQueueItem) pool to avoid memory allocations.  Similar to NetworkBuffers, one could just "get" an already existing struct from the pool.  This would need the ability to "grow the pool" on demand as well as the ability to specify how large you want the pool to be.  (just a thought on this)

ShadauxCat: I think this is a great idea!

text/0018-message-ordering-guarantee.md
```diff
@@ -0,0 +1,171 @@
...
+- `RpcQueueContainer.QueueItemType` now contains only three values: `ClientRpc`, `ServerRpc`, and `InternalCommand`. `CreateObject` and `DestroyObject` are removed, and are now a subcategory of `InternalCommand` <sup>[See Question #1]</sup>
+
+- `MessageQueueProcessor`'s (formerly `RpcQueueProcessor`) `m_InternalMLAPISendQueue` is removed, along with all functions in `MessageQueueProcessor` and `MessageQueueContainer` that reference it.
+
+- On the receiving end, the handling of all internal messages using the previous values from `NetworkConstants.cs` will instead handle a new `INTERNAL_COMMAND` constant, as it will need to pull the encapsulated message from the queue and process it before passing the wrapped content to the existing logic
+
+- Since structs can't have parameterless constructors, `NetworkUpdateStage` will be adjusted to appear as follows:
+  
+  ```C#
+  public enum NetworkUpdateStage : byte
```
NoelStephensUnity, line 115: Could you expand upon why something like this:
```
    public enum NetworkUpdateStage : byte
    {
        Initialization = 1,
        EarlyUpdate = 2,
        FixedUpdate = 3,
        PreUpdate = 4,
        Update = 0, // Default
        PreLateUpdate = 5,
        PostLateUpdate = 6
    }
```
 Would be different (as far as structs are concerned) than what you proposed and how adding the Unset value solves for the issue?  The original enum still has a "zero value" (i.e. default if not set), and so this area could potentially use some additional language as to why we need to make this change. 
(_mostly out of curiosity, but also to better clarify the justification_)

ShadauxCat: It's different because if Update is the 0 value, we wouldn't be able to tell the difference between "user didn't specify" (meaning we would have to do the smart logic to use the current stage) or "user explicitly requested this to run at the Update stage" (in which case we must not change it). I will update the RFC to be clearer on that.

ShadauxCat: Updated

text/0018-message-ordering-guarantee.md
ShadauxCat: Per our conversation - opted not to do any sort of caching as a part of this RFC as this RFC is primarily concerned with the sender side of things, and the primary benefit of caching would be on the receiver side. However, the sender is creating things it's not using, so to reduce allocations while I'm working on the sender side, I can create a new struct for send use that doesn't come with unused allocations. I've added that to the RFC.

text/0018-message-ordering-guarantee.md
MFatihMAR: +1 on this. if we'll provide a more intelligent default in the absence of a specific one, that'd be `Unset` not `Default` because we couldn't determine if it was really set or not otherwise. lgtm!

text/0018-message-ordering-guarantee.md
```diff
@@ -0,0 +1,175 @@
...
+This modification is fairly MLAPI- and Unity-specific. I'm unaware of any existing implementations that attempt to solve this particular problem.
+
+# Unresolved questions
+
+- Currently, there are two different message type values being passed through the system: the `RpcQueueContainer.QueueItemType` value and the value from `NetworkConstants.cs`. It appears the primary reason for this separation was to be able to identify which messages contained the queue item message structure, and which didn't, but is there still value to this separation? Or is there more value now in keeping only one and removing the other entirely?
+- Should internal commands that are triggered by users (such as `Spawn()`, `Despawn()`, and `ChangeOwnership()`) gain optional parameters to allow the user to control which stage they execute on?
+
+# Future possibilities
+
+As mentioned above in the rationale section, there is a known intention to integrate commands into the snapshot system. While this RFC doesn't attempt to address that yet, the consolidation of message sending mechanisms can be seen as a first step toward realizing that.
```
MFatihMAR, line 175: I think it's all good — I might try hard to nit-pick but this RFC, even though not being the ultimate solution to our queuing (or in wider terms, network input/output loop), it still fixes some issues and adds immediate value in the short term.

I might only suggest mentioning that all this work could be refactored or even potentially rewritten when we clearly define a network frame and network input/output loop. having said that, such changes would require other redesign/rework on the transport API abstraction and use of Unity.Collections more — just wanted to plug here instead of just saying LGTM 🚀 :)

NoelStephensUnity (approved):
Really nice work Jaedyn.

PR #19, opened 2021-06-15
Connect/disconnect Flow Updates

ShadauxCat:
[See the linked RFC document →](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/2e715860803e3a5f9b1101428c055e528528fd23/text/0000-disconnect-flow.md)

RFC proposing improvements to the user ability to react to connect and disconnect events, as well as exposing the existence of a client's peers and connect/disconnect events for peers.

### DIFF
--- a/text/0000-disconnect-flow.md
+++ b/text/0000-disconnect-flow.md
@@ -1,0 +1,134 @@
+- Feature Name: `disconnect_flow`
+- Start Date: 2021/06/02
+- RFC PR: 
+- Issue: [[Unity-Technologies/com.unity.multiplayer.mlapi#772](https://github.com/Unity-Technologies/com.unity.multiplayer.mlapi/issues/772)
+
+# Summary
+
+This feature standardizes the flow for communicating disconnects, both between server and client, and between SDK and user.
+
+# Motivation
+
+The motivation is to better enable users to react to network events, including servers reacting to disconnecting clients, clients reacting to lost connections to the server, and clients reacting to other clients joining and leaving the game.
+
+# Guide-level explanation
+
+When network topology changes (via new or lost connections, whether intentionally or unintentionally), it's often desirable for the processes that are (or were) involved in the network to be informed of it so that they can react. In order to reactto these events, MLAPI provides the following callbacks that can be registered to NetworkManager:
+
+- `OnClientConnectedCallback` - Called on the **server** when a new client connection is **formed**. Expects a delegate with the following signature:
+  
+  ```C#
+  void OnClientConnected(int clientId){}
+  ```
+
+- `OnClientDisconnectCallback` - Called on the **server** when a client connection is **lost**. Expects a delegate with the following signature:
+  
+  ```C#
+  void OnClientDisconnected(int clientId, NetworkManager.DisconnectReason reason)
+  ```
+
+- `OnPeerConnectedCallback` - Called on the **client** when another client has **connected** to the same server. Expects a delegate with the following signature:
+  
+  ```C#
+  void OnPeerConnected(int clientId){}
+  ```
+
+- `OnPeerDisconnectedCallback` - Called on the **client** when another client has **disconnected** from the same server. Expects a delegate with the following signature:
+  
+  ```C#
+  void OnPeerDisconnected(int clientId, NetworkManager.DisconnectReason reason)
+  ```
+
+- `OnServerConnectionEstablishedCallback` - Called on the **client** when a server connection is **established**. Expects a delegate with the following signature, where `peers` is the list of clients who are already in the game:
+  
+  ```C#
+  void OnServerConnectionEstablished(int[] peers){}
+  ```
+
+- `OnServerConnectionLostCallback` - Called on the **client** when it is **no longer connected** to the server (i.e., no longer a part of the game at all). Expects a delegate with the following signature:
+  
+  ```C#
+  void OnServerConnectionLost(NetworkManager.DisconnectReason reason)
+  ```
+
+The `reason` parameter is an enum currently defined as follows:
+
+```C#
+public enum DisconnectReason
+{
+    TimeOut, // Connection to the remote process timed out
+    Stopped, // The remote process shut down its network cleanly
+    Kicked, // The client was kicked from the server
+    ConfigurationMismatch, // The client and server have mismatched configurations
+    NotAuthorized, // The client is not (or is no longer) authorized to be connected
+}
+```
+
+# Reference-level explanation
+
+There are basically three components to this implementation.
+
+The first component is to enable to server to broadcast to clients which other clients are connected. This is implemented with two commands: a new `CLIENT_CONNECTED` command and the existing `CONNECTION_APPROVED` command. When a new client connects, a `CLIENT_CONNECTED` message with that clients ID is broadcast to all other connected clients to inform them of the new peer, and when the `CONNECTION_APPROVED` command is sent to the new peer, a list of existing connected client IDs is included. 
+
+ Upon receipt of one of these messages:
+
+- If the message is `CLIENT_CONNECTED`, the `OnPeerConnectedCallback` is invoked.
+
+- If the message is `CONNECTION_APPROVED`, the `OnServerConnectionEstablished` callback is invoked. **This is a change: The client will no longer receive the OnClientConnected callback, which is now server-only.**
+
+The other new message that's added is a `DISCONNECT` message, which contains in it the `DisconnectReason` value and the disconnecting client's clientId. A `DISCONNECT` message is sent:
+
+- From client to server any time `NetworkManager.StopClient()` or `NetworkManager.StopHost()` is called, providing `DisconnectReason.ShutDown` as the reason. **Server does not honor or use the clientId value for security reasons. It infers this value from the identity of the sender.**
+
+- From server to all clients with the `ServerClientId` any time `NetworkManager.StopServer()` or `NetworkManager.StopHost()` is called, providing `DisconnectReason.ShutDown` as the reason
+
+- From server to all clients with the disconnecting client's clientId any time it receives a `DISCONNECT` from a client, forwarding the received reason
+
+- From server to all clients with the disconnecting client's clientId any time connection to a client times out, providing `DisconnectReason.TimedOut` as the reason
+
+- From server to a client that has sent a `CONNECTION_REQUEST` message with a mismatched configuration, providing `DisconnectReason.ConfigurationMismatch` as the reason
+
+- From server to a client that has sent a `CONNECTION_REQUEST` message that's rejected by the `ConnectionApprovalCallback`, providing `DisconnectReason.NotAuthorized` as the reason
+
+- From server to all clients any time the server closes an established client's connection for any reason, providing `DisconnectReason.Kicked` as the reason
+
+When a `DISCONNECT` message is received:
+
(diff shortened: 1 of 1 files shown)


text/0000-disconnect-flow.md
```diff
@@ -0,0 +1,134 @@
...
+- 
+
+# Unresolved questions
+
+- Knowing that we can't update ConnectedClients and ConnectedClientsList, should we have a `HashSet<ulong> PeerList` available to users so they don't have to maintain such a list themselves?
+- Is there any possibility that providing a client with a list of anonymous client IDs for other clients poses a security risk? It's not clear what they could do with that information that would pose such a risk, especially since any objects owned by other clients will expose that same client ID.
+
+# Future possibilities
+
+There is the possibility additional disconnect reasons may be added in the future, or even the possibility we may want to provide the ability for the user to provide their own disconnect reasons. A `NetworkManager.Kick(clientId, reason)` method seems valuable. In that situation, we may consider making the reason a ulong with reserved values so users can add additional disconnect codes, or possibly even a string.
```
SamuelBellomo, line 134: It'd be awesome if we could have the ability to provide a payload with our responses.
Right now in Boss Room, we have this very ugly workaround with a custom message to give clients a reason for disconnection.
https://github.com/Unity-Technologies/com.unity.multiplayer.samples.coop/blob/9960a59586a521cf15cb8b477b2aa1c7370f3e13/Assets/BossRoom/Scripts/Server/Net/ServerGameNetPortal.cs#L292
https://github.com/Unity-Technologies/com.unity.multiplayer.samples.coop/blob/9960a59586a521cf15cb8b477b2aa1c7370f3e13/Assets/BossRoom/Scripts/Shared/Net/GameNetPortal.cs#L255

Do you think this future possibility will be in for our fall release?

text/0000-disconnect-flow.md
```diff
@@ -0,0 +1,134 @@
...
+- `OnServerConnectionLostCallback` - Called on the **client** when it is **no longer connected** to the server (i.e., no longer a part of the game at all). Expects a delegate with the following signature:
+  
+  ```C#
+  void OnServerConnectionLost(NetworkManager.DisconnectReason reason)
+  ```
+
+The `reason` parameter is an enum currently defined as follows:
+
+```C#
+public enum DisconnectReason
```
SamuelBellomo, line 57: I'm wondering if we could get inspired by HTTP error codes here.
for example, they have 4xx error codes which are client errors (asking for the wrong things or the wrong way) --> 400 is malformed request, 401 is unauthorized (where auth failed), 403 is when auth worked, but the user is forbidden from accessing, 404 for ressource not found, etc.
Same for 5xx error codes. We could have 500 for a reason where the server had an exception and the client couldn't connect (if there's any exceptions thrown server side during connection approval for example). Or 503 when the server was supposed to be there, but unavailable, signalling to the client to try again later (maybe using exponential backoff for example)
I don't think they are all applicable, but they might be a great inspiration for an exhaustive list of reasons?

text/0000-disconnect-flow.md
```diff
@@ -0,0 +1,134 @@
...
+- `OnClientDisconnectCallback` - Called on the **server** when a client connection is **lost**. Expects a delegate with the following signature:
+  
+  ```C#
+  void OnClientDisconnected(int clientId, NetworkManager.DisconnectReason reason)
+  ```
+
+- `OnPeerConnectedCallback` - Called on the **client** when another client has **connected** to the same server. Expects a delegate with the following signature:
+  
+  ```C#
+  void OnPeerConnected(int clientId){}
```
SamuelBellomo, line 33: I love we're making the difference between client and peer. That's something that was pretty confusing the first time I looked into this

text/0000-disconnect-flow.md
ShadauxCat: I don't think it'd be difficult to add this for the fall release. I was trying to keep this RFC a little slimmer and not try to spec out the ways a client could get willfully disconnected from the server, but if it's valuable to add now, I'm on board with adding that functionality to this RFC. I can't speak to scheduling of implementation though, of course. :)

I'll update the RFC to add the ability for the disconnect to carry a custom UserData buffer that can be passed opaquely into the callback. Do you think that would solve your need here? Something like `void Kick(ulong clientId, DisconnectReason reason, NetworkBuffer userDataBuffer = null)` - or is there a way that the third parameter could be an `INetworkSerializable` and get properly constructed on the other end without the user having to deal with buffers directly? (I'm not sure if INetworkSerializables pass any type information with them that could be used to construct it on the other side in a generic way or not.)

text/0000-disconnect-flow.md
```diff
@@ -0,0 +1,134 @@
...
+
+- From server to all clients with the `ServerClientId` any time `NetworkManager.StopServer()` or `NetworkManager.StopHost()` is called, providing `DisconnectReason.ShutDown` as the reason
+
+- From server to all clients with the disconnecting client's clientId any time it receives a `DISCONNECT` from a client, forwarding the received reason
+
+- From server to all clients with the disconnecting client's clientId any time connection to a client times out, providing `DisconnectReason.TimedOut` as the reason
+
+- From server to a client that has sent a `CONNECTION_REQUEST` message with a mismatched configuration, providing `DisconnectReason.ConfigurationMismatch` as the reason
+
+- From server to a client that has sent a `CONNECTION_REQUEST` message that's rejected by the `ConnectionApprovalCallback`, providing `DisconnectReason.NotAuthorized` as the reason
```
SamuelBellomo, line 91: Connection approval isn't necessarily about authorization though. You could have a game check to make sure you don't have too many players of a certain type or coming from a certain region (for whatever reason). That wouldn't be "authorization"

text/0000-disconnect-flow.md
ShadauxCat: I think that's valuable. I think I would prefer, personally, to still see them represented as an enum, but having the enum values logically grouped would help with debugging, and we could follow the same prefix as HTTP just to make it more familiar. 400-499 is something wrong with the request (not authorized, malformed connect request, version mismatch, etc), 500-599 is something wrong with the request handler (unhandled exception in authorization callback, time out, etc)

text/0000-disconnect-flow.md
ShadauxCat: True, in line with the comment about HTTP-based error codes above, we can change the connection approval handler to return a reason for not approving. Possibly also integrate the user data as well with that.

text/0000-disconnect-flow.md
SamuelBellomo: oh you're right, I wasn't suggesting we use ints for these error codes. I don't really mind ints vs enums.
It's mostly about the list of errors we can get inspiration from. HTTP has been established for a while, it has a pretty good list of use cases we might want to express too.
Good idea on the user extensible values.

text/0000-disconnect-flow.md
SamuelBellomo: ok ok, it's really up to you. I just wanted to make sure this didn't fall into cracks :)
we have multiple ways to handle that serialization (that we really should make unique) You can look into how netvars currently do it. Thing is, there's plans with snapshot to look at serialization, you might want to have a quick sync with @jeffreyrainy and @LukeStampfli 
Do we need type information though? Both client and server would be implemented by the same team, the person implementing the client would know what data type the server sends in their own custom payload and would be able to deserialize using the appropriate type, no?

ShadauxCat: The question I had here was, could we make something like

```OnPeerDisconnectedCallback(DisconnectReason reason, INetworkSerializable payload)```

And have `payload` get deserialized 'magically' by the SDK code, which wouldn't know what type `payload` is expected to be? I'm guessing probably not - which is why I proposed using something like `NetworkBuffer`, but I do feel that `NetworkBuffer` is a bit more inaccessible to less experienced teams, so if there was a way to make the payload magically deserialize to the correct type, it'd be a better developer experience.

In this case, unlike RPCs and NetVars, we don't necessarily have any shared understanding on both sides at compile time of what that type will be, due to the generic nature of the callback type signature.

Comment (LukeStampfli) [reply text/0000-disconnect-flow.md:134]: We currently do not support typed serialization and I don't think it is a feature which we can just quickly add to MLAPI.

What we could do is use generics for this. The API would then look like the following:
```csharp
OnPeerDisconnectedCallback(DisconnectReason reason, Payload payload)
```

`Payload` would be a struct which a) exposes a `NetworkBuffer` property for direct network buffer access and b) it will have a `As<T>` function to convert the payload automagically into the right type. So for instance:
`payload.As<MyCustomSerializable>();` would extract the payload as a `MyCustomSerializable`.

The only restriction here would be that the sides writing and reading the payload must use the same generic type. But I'd argue that is something that we want because it is much cleaner then typed serialization.

text/0000-disconnect-flow.md
```diff
@@ -0,0 +1,134 @@
...
+  void OnClientConnected(int clientId){}
+  ```
+
+- `OnClientDisconnectCallback` - Called on the **server** when a client connection is **lost**. Expects a delegate with the following signature:
+  
+  ```C#
+  void OnClientDisconnected(int clientId, NetworkManager.DisconnectReason reason)
+  ```
+
+- `OnPeerConnectedCallback` - Called on the **client** when another client has **connected** to the same server. Expects a delegate with the following signature:
```
LukeStampfli, line 30: I'm not sure whether I like the the OnPeerConnected/Disconnected callbacks. The reason for this is because it breaks the enclosure principles of start topology (a client only knows about the server but not about other clients). While for many games this might be fine, I can see some cases where this would be not desired, that's why I think we should consider to not add this as a core library feature. The problems I see with this are:

- In some larger scale games like Battle Royale or MMOs this would not be desired because a) it adds more bandwidht. b) it allows for cheats / exploits.
- This is not a complete feature. If a new client joins it won't know the list of peers which are already connected. It will just get events for newly connecting/disconnecting peers.

I suggest that if we want to implement this for now, we would rather do it in a high level component (NetworkBehaviour). I don't see a good reason why that's not the better approach. We can do something along the lines of:
```csharp
public class PeerConnectionHandler : NetworkBehaviour
{
    public OnPeerConnect OnPeerConnectCallback;

    private override void NetworkStart()
    {
        if(IsServer)
        {
            NetworkManager.ClientConnectedCallback += OnClientConnected
        }
    }

    private void OnClientConnected(int clientId)
    {
        Assert.IsTrue(IsServer)
        OnClientConnectedClientRpc(clientId);
    }
    
    [ClientRpc]
    private void OnClientConnectedClientRpc(int clientId)
    {
        OnPeerConnectCallback.Invoke(clientId);
    }
}
```
And then can tell people to drop this into their scene on a NetworkObject if they want peer connected events.

ShadauxCat: > This is not a complete feature. If a new client joins it won't know the list of peers which are already connected. It will just get events for newly connecting/disconnecting peers.

Actually the RFC does account for this:

> When a new client connects, a CLIENT_CONNECTED message with that clients ID is broadcast to all other connected clients to inform them of the new peer, and when the CONNECTION_APPROVED command is sent to the new peer, a list of existing connected client IDs is included.

Though I can see how that's easy to overlook, and I did neglect to include a parameter in the callback to actually provide that information to the user.

I think the idea of putting this in an optional NetworkBehaviour is good though.

text/0000-disconnect-flow.md
ShadauxCat: 100% agree, typed serialization is a little scary. Generics is a better option, gets the ability to use INetworkSerializables so it's accessible but it doesn't add the bandwidth and CPU overhead of sending and decoding type info.

PR #22, opened 2021-07-06
Update RFC Template and Reformat Existing RFCs

MFatihMAR:
This PR is consists of 2 main changes:

1. Updating RFC template where "Feature Name: XXX" field becomes a main title on top of the RFC document so that feature name is obvious and clearly visible for reading.
2. Replacing "Issue: #0000" with "SDK PR: #0000" and getting rid of intermediary tracking issue concept. This is suggested because originally tracking issue was aiming to cover scenarios where an RFC would be implemented by multiple PRs instead of just one. However, based on the history record so far, we tend to implement an RFC in one PR with all the features and tests. (Having said that, we could still put multiple PR links next to each other like `SDK PR: #123 & #789` etc.)


LukeStampfli (approved):
Looks good to me. Getting rid of tracking issues makes sense for me as it is unknown when we will have time to automate this and it reduces RFC overhead 👍

PR #23, opened 2021-07-08
RTT computation

jeffreyrainy:


SamuelBellomo (approved):

andrews-unity:
So my first question is why do we need an overly generic high-level solution? I mean you brought up "snapshotting, interpolation, and client-side prediction" but in our case some of those systems could use values specific to the snapshotting system? I mean do you see any reason why we would want to encourage devs not to use what we provide? 

I think if you look at a lot of other system out there (even say DOTS Netcode) they kind of go with a specific solution for their snapshotting paths. I guess what I am getting at is that I am not sure we actually need a per connection level API exposed. One of the traps I think we get into is trying to go wide and provide a solution for all use cases but really we should try to support our use cases the best and the framework is an example to those who want to change things if they so choose to.

jeffreyrainy:
> So my first question is why do we need an overly generic high-level solution? 
> I mean do you see any reason why we would want to encourage devs not to use what we provide?

I'm very glad you raise those points.

On the first point, the initial suggestion I had made was to do exactly that. Internal to Snapshot System. Simpler, not overly generic. I'll discuss details off-line with you.

Agreed on the second point too, but just to clarify: providing the notification for sends and acks was still `internal`, so it would not "encourage game devs to not use what we provide" it would merely "encourage our own team to use various different RTT sources", which is also a valid point. </nit> :-)

text/0000-rtt-computation.md
```diff
@@ -0,0 +1,119 @@
...
+- `NotifySend(45, 1001.99);` when it sends message 45.
+- `NotifyAck(43, 1002.01);` when it receives a message acknowledging 43.
+
+At this point, if game code were to call `GetRtt()`, they would get
+- best RTT of 0.15
+- worst RTT of 0.54
+- average RTT of 0.345 
+- sampleCount of 2
+
+Notice that unacknowledged messages don't contribute to RTT computation. Neither do lost messages. Messages that ended up being retransmitted by lower-level transport would end-up with higher measured latency. This is by design and a feature. From game code point-of-view, this latency exists and affects gameplay. As such, it must be measured and reported. It is another way in which transport RTT and high-level RTT differ.
```
LukeStampfli, line 76: > Messages that ended up being retransmitted by lower-level transport would end-up with higher measured latency.

Do we plan to do this or not? Right now my understanding is that the lower-level transport will not retransmit messages?

text/0000-rtt-computation.md
```diff
@@ -0,0 +1,119 @@
...
+public struct Rtt
+{
+    public double best;
+    public double average;
+    public double worst;
+    public int sampleCount;
+}
+```
+
+The members of this struct describe, for a given client, the RTT. `best` shows the best RTT measured, `average` shows the average RTT measured, and `worst` shows the worst RTT measured. `sampleCount` indicates how many samples were used to get those values. A count of 0 indicates no data is available. In this case, the other fields would be 0.
```
LukeStampfli, line 37: For average we could consider using an exponential moving average which usually gives a better result.

text/0000-rtt-computation.md
jeffreyrainy: This RFC describes a modular system for RTT computation that is not tied to the source of messages/acks. So you can't say "lower-level transport will not retransmit messages". We don't know. Some part of MLAPI could use reliable messages and get hidden retransmits and compute RTT from it. Some other part of MLAPI could use unreliable message and then this would not apply.

I think the argument you just made is one for not having modular RTT. If the fact that snapshot is not retransmitting message is a relevant detail, then RTT could have been implemented in the snapshot system in the first place.

text/0000-rtt-computation.md
jeffreyrainy: Yes. Good idea. I would just argue, then, to add a `internal void Reset()` to `NetworkConnection`, because exponential moving average keeps contributions forever (with decreasing weight, but still).

text/0000-rtt-computation.md
LukeStampfli: I do not understand. Would we do sequence ack on top of a ReliableSequenced protocol? Why? I think we should rather  be opinionated here on how this should be used.

jeffreyrainy: The simplest resolution would be to remove the text "Messages that ended up being retransmitted ...". I'm OK with that.
 
 Initially, I wanted the RTT computation to be in snapshot system and not exposed. I would still be OK with that, and it would also resolve the issue.
 
The reason for this text in the RFC is that you asked for the RTT be put outside of Snapshot. Once RTT computation is put outside, we get to rely on the using system to provide Send and ACK times. By making it generic and not knowing what the using system is, we get to also not know whether this system uses reliable or unreliable messaging.

If this system uses unreliable messaging, then, this question is resolved. The text doesn't apply.
 
If this system uses reliable messaging (even though it might be a bad idea), then the question arises: what does it do the the RTT computation. What this aims to state is that it is still fine. If your system uses reliable messaging and you feed the RTT computation system with your Send and ACKs time, the RTT you get will be correct on average because the computation will be affected in the same way as your payload. The best and worse RTT will also be valid.

andrews-unity:
So this is what I would like to purpose for now... I think we should limit this system to the Snapshot system, right now this is really the only system that requires this kind of thing and I don't see a reason to build something super general for use cases we haven't really come across and like other systems I think if users want to roll their own they will have a great reference via the snapshot system. I think keeping it there makes things less complicated, it keeps the code contained but if written in such a way could be abstracted out later if we eventually come across a need for it.

text/0000-rtt-computation.md
LukeStampfli: I was not trying to say we make RTT generic. In my mind Snapshots have nothing to do with RTT, the sequence acknowledge protocol of our messaging system does. So I was trying to ask why this is part of the snapshot system and if there is no reason for having it in there I suggested moving it somewhere else.

As for exposing RTT the obvious choice in my opinion is to use whatever represents a peer which in our case seems to be `NetworkClient` (or `NetworkConnection` soon).

LukeStampfli:
I don't see why RTT would be tied to the snapshot system. We need access to RTT in other fundamental pieces of MLAPI such as the time system and I can think of some future systems which will need access to this too such as inputs/commands.

I can see why we have RTT related to a sequence ack algorithm or maybe to ticks. I don't see why it would be related to snapshotting.

text/0000-rtt-computation.md
jeffreyrainy: > I don't see why RTT would be tied to the snapshot system. We need access to RTT in other fundamental pieces of MLAPI such as the time system and I can think of some future systems which will need access to this too such as inputs/commands.
> 
> I can see why we have RTT related to a sequence ack algorithm or maybe to ticks. I don't see why it would be related to snapshotting.
> 
> By creating a dependency to the snapshot system for RTT we end up with a circular dependency of tick > snapshots > RTT > tick
> which means we will ever only be able to use or test all of them and changes to each of them will need to consider every other system. Essentially it will become one monolithic design where everything depends on everything else.

I'm also OK with it in `NetworkConnection`. The `notify...` API are internal. At first they will only be used by Snapshot, but the decoupling is there.

andrews-unity: So I think having a NetworkConnection abstraction, while we force all Transports to have some sort of connection tracking etc... is kind of gross. So if we are going to introduce the idea of a NetworkConnection I would like to ensure that we have something that at least can be used consistently across the SDK as a whole including the Transport API.

jeffreyrainy: I'm not suggesting a new abstraction. `NetworkConnection` is just a rename of the existing `com.unity.multiplayer.mlapi/Runtime/Connection/NetworkClient.cs`. There's a shortcoming that this is present only for clients, and not for the server. RTT needs to be stored for both clients and servers.

If we keep RTT internal to Snapshot System, I'll have a struct per client anyway on the server, and one for the server on client machines.

If we keep it out of the Snapshot System, the existing `NetworkClient` is a reasonable place. But we just need to add it for the server, too.

Are you suggesting we remove from mlapi the existing `NetworkClient` ?

andrews-unity: I think the renaming NetworkClient->NetworkConnection is really outside the scope of the conversation to be honest. So I think we should keep focus on should the RTT be in the Snapshot System or should it be stored somewhere outside so it could potentially be used by one system and potentially future systems. 

I get your point @LukeStampfli with regards to sequence and RTT not being a part of snapshots... my main concern about exposing this API beyond Snapshots is outside the Tick system... there are no other systems that we have today that need this API and yes I know future things could use it but right now we need to make progress on this system and get it done. 

My second concern is lets say for the sake of argument we store it outside of Snapshots... My main concern is now we have this RTT system per object thats driven by something other than the Transport RTT which we say isn't right for this use case and I think that leads to confusion and inconsistent APIs which is why I mentioned above that if we are going to store it outside of the Snapshot then lets reconcile the idea of what RTT is for the SDK through and through and not pidgin hole it to these specific use cases.

LukeStampfli: I think the way I envisioned this to work would be:

1. Our transfer (message) system has a way to send messages over a seq-ack protocol. It keeps track of sent seqs and when an ack arrives calculates an RTT.
2. Once per tick some system (probably not snapshot system) creates a snapshot message based of the data of the AOI and snapshot system and sends that over the transfer system. But the transfer system can support any kind of messages not just snapshot.
3. `NetworkConnection`/`NetworkPeer` or something like that has a way to get updated with an RTT value. When it gets updated with an RTT value it will calculate some kind of average which it then exposes as a property.

Now we can decide which messages from 1. pass their RTT into 3. We have a system which is completely decoupled. 3) can exist without 1) or 2). 1) can exist without 3).

As for having different RTT values. I'd argue all our systems should care about the same value which is the time an unreliable packet takes from flush -> transfer -> poll -> flush -> transfer -> poll. 

But if we don't go for a design like that I guess I'm fine just putting it all into this one Snapshot System"

andrews-unity: A couple questions... I assume 1 doesn't exist yet and so that would have to be done as part of this work ? I am asking because I have not seem the term transfer system thrown around before :P 

For 3 I would not be against something like a NetworkPeer if we could consolidate this into the transport API or at least remove RTT from the transport API all together and provide some API to access this data? Is there any reason we can't store the NetworkPeer data on the Transport layer? (I am not saying we should I am just asking)

LukeStampfli: Yes 1. doesn't exist yet. What I mean by 'transfer' is a high level protocol system (think like a reliability stage which is aware of its content). We'll have a bunch of systems which each will need to send data in various manners (unreliable, pessimistic (all cost delivery, reliable-sequenced) etc. (there are a bunch of approaches to this like delta snapshotting, eventual consistency and each of them can be implemented with different priority/interest systems) What I'm trying to advocate for, for a while is that we build something here which is not coupled to snapshots. Because I think in the long term snapshots is just one of the many things which we will need to transfer over the network.

I think having RTT in the transport API is a great approach if we have one highly optimized and tightly coupled transport to MLAPI. From what I know this is quite common and competitors chose this solution on purpose because of all the upsides. Right now with MLAPI we are much more open at supporting different kinds of transports which is limiting but provides us with other upsides. But it kind of forces us to move a lot of things out of the transport layer into the higher level.

jeffreyrainy: > From what I know this is quite common and competitors chose this solution on purpose because of all the upsides.

Personally, what worries me about this stance is that mlapi runs as an external package, driven by the Update() frame rate. This makes it so that our non-threaded networking, or at least the high-level part of it, doesn't get to run in the background. As a result, knowing that transport has a RTT of say, 5 ms, is useless if I'm going to wait 10 ms in order for the frame to finish (100fps) before Snapshot gets to run. That 5ms doesn't help me with client-side lag compensation, because there's a higher latency cost hidden.

Now, if we were to refactor it all and have threads running in the background and Snapshot running at a rate decoupled from game updates, then I'd be 100% in agreement with you. However, I don't think we are there yet and I fear mlapi as a package of Unity as currently designed would actually prevent this "run in the background" architecture.  It's not even possible to touch game object outside of Update() (and matching delayed FixedUpdate())

MFatihMAR: > It's not even possible to touch game object outside of Update() (and matching delayed FixedUpdate())

Once we have `NetworkTickSystem` and `NetworkTimeSystem` in-place, we'd probably rely on `NetworkFixedUpdate()` rather than `MonoBehaviour`'s `Update()` or `FixedUpdate()` and I'd argue `SnapshotSystem` to implement its logic under `NetworkFixedUpdate()` executed by `NetworkTickSystem`+`NetworkTimeSystem` so that `SnapshotSystem` would be integrated with ticks, time and potentially RTT in the future.

In terms of transport RTT vs framework RTT, I think RTT should be calculated based on when the tick advances in a given so-called NetworkFrame/NetworkTick. If the time progresses and `NetworkTickSystem` executes `NetworkFixedUpdate()`, RTT should also be calculated based on `current time - sent time` because that's when the data becomes available to RTT-sensitive systems. Let's say we sent a snapshot, we should calculate RTT around network ticks/updates rather than transport ACKs. Just my 2-cents though, I haven't thought about all different features and use cases involved / will potentially involve.

MFatihMAR:
I think I agree with @LukeStampfli here on pointing the difference between transport RTT vs framework RTT and the fact that calculation (probably have to be) should be done at the framework-level for framework-level features such as snapshots, time etc. We will define our very own `NetworkTickSystem` and `NetworkTimeSystem` which would form basics of our "network frame" and framework-level RTT should be calculated around that since systems in the framework should also be tied to network tick/frame/update/time etc. A new framework feature (such as snapshot, networktransform or whatsoever) shouldn't arbitrarily do its update on `MonoBehaviour.Update()` or `NetworkBehaviour.FixedUpdate()` and rather do its update on `NetworkTickSystem.NetworkFixedUpdate()` to make sure it aligns well with other systems and again, forms a "network frame". In that regard, framework-level (or framework-feature-level) RTT should be calculated on the framework-level, not transport-level.

Having said that, just for the sake of simplicity and getting some stuff done, I'd also agree with @andrews-unity here to get something working for snapshots, just for now and later in the future, further investigate/prototype/iterate on this framework-level RTT stuff. I can see RTT being discussed in its own RFC again once we settle `NetworkTickSystem`, `NetworkTimeSystem`, "network frame" and tick/time/frame dependent systems integrated. RTT could be moved out of snapshot system and outsourced to that new "RTT system" that sits on top of `NetworkUpdateLoop` + `NetworkTickSystem` + `NetworkTickSystem` + "Network Frame (?)".

text/0000-rtt-computation.md
andrews-unity: > It's not even possible to touch game object outside of Update() (and matching delayed FixedUpdate())

This is not entirely true :P you can touch them outside those areas you just need to be mindful as to what you are doing. 

I don't think that I am saying Transport owns the calculating etc... of the RTT I think there should just be one RTT :P and not two values that confuse people as to which one to use right now we have a Transport RTT which to be honest is pretty useless, and I am going to guess un-used. If we say hey lets have this higher level RTT what prevents us from at least storing this data in the transport or somewhere there about? as it just logically makes sense that Peer data be stored there.

As for being driven by update? I don't think running as a package prevents us from "Run in the background" at all, here again as I mentioned before you can touch GameObjects from outside the main playerLoop functions we just need to be mindful of when and where we are doing it. Also we can touch our own data as long as we know it to be safe, and there are ways to be able to actually update transforms from jobs but it's complicated. So I think we can get there even piece meal.

MFatihMAR: I think I agree with Andrew here. Let's have one framework RTT and no transport RTT (with good `<summary>` blocks and documentation explaining how it's being calculated).

jeffreyrainy:
Given Andrew's and Fatih's feedback, for now, I'll put the RTT computation in snapshot. I'll aim for ease of moving it, later, but for now it should have less external impact.

Thanks for the good feedback. I'll close this.

text/0000-rtt-computation.md
mattwalsh-unity: I was going to throw out a pipe bomb so suggest `NetworkManager.ConnectedClientsList[index].GetRtt();` return a collection of results, one per channel.  Then the specialized snapshot ACK mask-based RTT can be in there for a snapshot channel, but then other channels (or QoS's, if we finally devolve channels to be QoS) they can have their own RTT value, which could be computed in MLAPI or reported back from the transport, etc.

I mean for something as simple as a game wanting to show ping time, what would our developer guidance be?  Just use the snapshot RTT?

PR #25, opened 2021-07-14
Rfc/interest management

mattwalsh-unity:
At long last, here's my RFC draft


text/0015-interest-management.md
```diff
@@ -0,0 +1,258 @@
...
+```
+
+As we described earlier, this node maintains its own storage for the objects it tracks (`ManagedObjects`).  When it is told one of its associated objects is added, it simply adds it to its storage and vice versa when an object is going to be deleted.  At query time, if we have no kernels we simply merge the results passed in from the other clients with all the objects we manage and do no other computation.  Otherwise we invoke our kernel(s) and merge in those results.
+
+A default instance of this node, by the way, is instantiated in the `NetworkingManager` and is where objects go that have no associated `InterestNodes`.
+
+Now let's author the `RadiusInterestKernel`.  Actually, this file is included but let's look at its implementation:
+
+```
+public class RadiusInterestKernel : InterestKernel
```
LukeStampfli, line 126: I'm trying to understand how `InterestKernels` would be used in a real world example. A lot of common approaches to AOI (spatial hashing, octrees, radix sort etc.) are stateful and calculate their state in a single pass over all objects, so the QueryFor interface of the `InterestKernel` won't really work for them. So from what I understand `InterestNode` would have to be used instead of `InterestKernel`, is that correct?

In the description about InterestStaticNode above it is mentioned that "for a given prefab we want to consider it relevant if it is within a given radius and on the same team as the querying connection's object" would be a use case. 
But from my understand this kernel approach is an `OR` and not an `AND` operation. So instead of `all objects within a given radius AND on the same team` we query for `all objects within a given radius OR in the same team`.

text/0015-interest-management.md
```diff
@@ -0,0 +1,258 @@
...
+* We want the aforementioned schemes to be composable.  For instance, a user should be able to have both a AOI scheme and a "always synchronize players who are in the same cohort" scheme and have the results from each efficiently combine
+* Each scheme should be free to decide how to implement relationships; so long as the scheme can take in a client and return a set of objects it should be free to come up with that answer using whatever storage and computation it wants
+* Each scheme can therefore decide at what cadence it does its work.  A scheme can choose to dynamically come up with an answer every frame.  Or, it could return cached results each time except every 30th frame.
+* We also want a means for code outside the scheme to be able to trigger an update.  That is, there are times when it's best for the scheme itself to know when it's time to do things like the cache/compute decision, but then there are other times when game code outside the Interest System knows better that it's time for the scheme to adjust itself
+
+# Guide-Level Explanation
```
LukeStampfli, line 33: So there is just a `guide` level section and not `reference` level section currently. But the below looks to me more like a `reference` section maybe?

I think the technical details are well covered in this RFC. What I think is missing is how we expect a user to use this.

As a beginner how would approach this AOI system?
- Use a bunch of pre existing scriptable object InterestNodes we provide and combine them in a config to get my desired result.
- Maybe use an InterestKernel for custom AOI  code because they have a more beginner friendly interface then the Interest node?

As an advanced user I guess I would end up doing the same as a beginner but if I have special performance or gameplay needs I would write a custom InterestNode?

Or is this RFC more targeted towards us MLAPI developers and we will look into the UX later?

mattwalsh-unity: Yeah, I at first was going to try and divide into guide and reference, but the natural fit seemed to me to explain how it works in an example using it.  I just updated it with some screenshots that I think should make it a bit more consumable

text/0015-interest-management.md
mattwalsh-unity: Ok, good insight.  It is true, you would not likely do a domain decomposition by just standing up an `InterestKernel` and sticking it on a `InterestNodeStatic`.  However, like in the odds/evens example, what I would hope you would do is put the 'dispatch to the appropriate grid node' in a custom `InterestNode`.  I mean, that has to be custom.  But then, that node would then dispatch the correct grid node which (I think) would be `InterestNodeStatic` instances.  And if you wanted to do processing beyond just 'return all the things in this grid cell', then you'd attach an `InterestKernel` to them.   If not, just like the odds/evens example, the grid cell has no interest kernel and just returns everything.

mattwalsh-unity: On your second comment, thank you, good point, I'll correct it, it is indeed an `OR` relationship and I meant it to be (e.g. I want to see my team members regardless if they're close by).  

But the next question I have for you & the internet, and @TwoTenPvP raised this to me in a 1:1, do we think it's important to let you define `AND` relationships?

text/0015-interest-management.md
```diff
@@ -0,0 +1,261 @@
...
+* Therefore, we desire an interface that takes in a connection and returns the objects for that client
+* We want the way this question is answered to be user-extensible to support all kinds of different scenarios.  Domain decomposition / geometric / AOI is very common, but we also want developers to be able to come up with creative, game-specific **schemes** that have nothing to do with spatial relationships
+* We want the aforementioned schemes to be composable.  For instance, a user should be able to have both a AOI scheme and a "always synchronize players who are in the same cohort" scheme and have the results from each efficiently combine
+* Each scheme should be free to decide how to implement relationships; so long as the scheme can take in a client and return a set of objects it should be free to come up with that answer using whatever storage and computation it wants
+* Each scheme can therefore decide at what cadence it does its work.  A scheme can choose to dynamically come up with an answer every frame.  Or, it could return cached results each time except every 30th frame.
```
NoelStephensUnity, line 31: Is the "every 30th frame" pertaining to the tick update?  What consumes every 30th frame?

text/0015-interest-management.md
```diff
@@ -0,0 +1,261 @@
...
+# FAQ
+
+## Why can one associate more than one `InterestNode` with a prefab?
+
+Let's imagine you're setting up your player's prefab.  Say you want players can see other players if they're some distance away using a dispatched / custom storage like the second example.  But let's say we also want to enable players to see other players who are on the same team, regardless of distance.  You could write a special `RadiusOrTeam` combination node that does both.  For better re-use, however, the system allows one to add more than one node; in this case you would have a `GraphInterestNode` *and* a `TeamInterestNode`.  Then, when the query happens on player prefabs the result from both nodes is merged
+
+
+
+# Drawbacks
+- The U/X is a bit cumbersome in the `InterestNodeStatic` case in that you must create a node, then create the kernel and then associate both to the prefab
```
NoelStephensUnity, line 246: This process can be automated.

text/0015-interest-management.md
```diff
@@ -0,0 +1,261 @@
...
+
+# Prior art
+
+# Unresolved questions
+- as an alternative to having storage of nodes under management in `InterestNodeStatic`, what about an API change to allow a node to simply get access to all spawned objects?
+- or, put another way, should the Interest System become the single source of truth for `SpawnManager.SpawnedObjects`?
```
NoelStephensUnity, line 252: This would make sense.  This way you could have a default radius for your query of spawned objects, but if needed you might need to perform predictive calculations that requires viewing "beyond your current" AOI radius.  This would be useful for late joining player synchronization (near future) and streaming worlds (etc).

NoelStephensUnity (approved):
This looks really good Matt.
I am interested to see how this evolves over time!

PR #26, opened 2021-07-21
Serialization RFC

ShadauxCat:
[See the linked RFC document →](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/serialization/text/0026-serialization.md)

RFC proposal detailing a high-performance, zero-allocation alternative to our current NetworkReader and NetworkWriter classes and their brethren.


text/0000-serialization.md
```diff
@@ -0,0 +1,394 @@
...
+    // Handle the message
+    void Handle(in NetworkContext context);
+}
+```
+
+Additionally, each message should have a method on it as follows:
+
+```
+[MessageHandler(IMessage.MessageType.MyMessageType)]
+static void Receive(ref FastNetworkReader reader, in NetworkContext context);
```
LukeStampfli, line 70: I do not understand why we need ILPP here for something like this. For Server/Client RPC we needed this because we wanted to rewrite how c# code behaves before we pass it to the compiler. Looks like this is just creates a lookup table from MessageType to Action<ref FastNetworkReader, in NetworkContext>.

We can just construct this lookup at runtime initialization with a Dictionary<MessageType, Action<...>>. This will result in roughly the same performance but reduces the amount of ILPP in our codebase.

bendoyon:
I'm a huge fan of this.

We were currently implementing something in the same vein for metrics. (We need to serialize the payload that gets displayed in the Unity profiler.)

With that in mind, I was wondering if it would be possible to this more generic (ie: probably just naming at this point, but to make FastNetworkWriter and FastNetworkReader into SomethingSomethingWriter and SomethingSomethingReader so they can be used in different contexts.)

We'd also use codegen for generating serializers for the different structs we use for tracking metrics.

text/0000-serialization.md
```diff
@@ -0,0 +1,394 @@
...
+
+- This approach achieves extremely fast performance
+- This approach avoids memory allocations at the core layer except by way of `NativeArray`'s `Allocator.Temp` and `Allocator.TempJob`, which are orders of magnitude faster, and are only used in situations where incoming data needs to be moved from memory we don't own to memory we do, or where we need a temporary buffer to avoid reserializing.
+- This approach makes our message formats clean, easy to understand, and easy to debug.
+- This approach gives each message its own "home" in the form of the struct that contains it, making it easy to find all the code that serializes, deserializes, and processes the message.
+- This approach serializes data exactly once, removing a current step wherein data is deserialized, changed, and reserialized
+
+Ultimately, this approach not only increases the performance of MLAPI, but also reduces the barrier of entry for new developers and increases the readability and transparency of our codebase - all of which will have the effect of increasing our users' ability to feel confident in MLAPI.
+
+# Prior art
```
LukeStampfli, line 374: There is some prior work here which could be explored (I think overall what's presented in this RFC is better than the prior work but I think it might still be interesting to take a look at it)
- [UNet NetworkMessages](https://docs.unity3d.com/Manual/UNetMessages.html)
- [Mirror NetworkMessages](https://github.com/vis2k/Mirror/blob/master/Assets/Mirror/Tests/Editor/NetworkMessageTests.cs)

text/0000-serialization.md
ShadauxCat: We may not need ILPP. My reason for proposing it is that I was concerned about iOS, since they don't allow the use of JIT. My concern is that if we never call these functions directly and we only discover them at runtime via reflection, that the AOT compiler would determine they're unused and not emit code for them, and then things would crash on iOS when we tried to call those functions and the JIT compiler wasn't available to compile them.

This may be an entirely invalid concern, in which case, yes, it would be much simpler to discover them at runtime.

text/0000-serialization.md
ShadauxCat: Yep, was planning on looking at those, as well as looking more at DOTS netcode, but wanted to get some eyes on the current proposal while I look at them.

ShadauxCat:
@bendoyon 

> With that in mind, I was wondering if it would be possible to this more generic (ie: probably just naming at this point, but to make FastNetworkWriter and FastNetworkReader into SomethingSomethingWriter and SomethingSomethingReader so they can be used in different contexts.)

FastBufferWriter/FastBufferReader?

text/0026-serialization.md
```diff
@@ -0,0 +1,391 @@
...
+{
+    private float f;
+    private bool b;
+    private int i;
+
+    void Serialize(ref FastBufferWriter writer);
+}
+```
+
+Serialize can be implemented in two ways:
```
LukeStampfli, line 143: I really like this approach where we can either basically do a memory copy or a packed write based on the message type.

bendoyon:
Are we considering implementing some or all of this in a separate package?

ShadauxCat:
This RFC is now finalized (pending comments an revisions) and ready for review.

ShadauxCat:
@bendoyon I hadn't considered putting anything in a separate package. I can see the benefit, but I don't know if I'm able to make that call or if that's more of a Jesse/Matt/Gio call.

bendoyon:
> 
> 
> @bendoyon I hadn't considered putting anything in a separate package. I can see the benefit, but I don't know if I'm able to make that call or if that's more of a Jesse/Matt/Gio call.

@JesseOlmer @mattwalsh-unity @UnityGio 

Would be useful so the tools package can then leverage it, but I understand it might add some complexity / time.

JesseOlmer:
> > @bendoyon I hadn't considered putting anything in a separate package. I can see the benefit, but I don't know if I'm able to make that call or if that's more of a Jesse/Matt/Gio call.
> 
> @JesseOlmer @mattwalsh-unity @UnityGio
> 
> Would be useful so the tools package can then leverage it, but I understand it might add some complexity / time.

I think Andrew already spoke to this a bit in slack - it's something we could consider in Collections at some point, but right now is not the appropriate time. A new package has an extremely high process overhead so that's probably not the right direction.

text/0026-serialization.md
```diff
@@ -0,0 +1,598 @@
...
+        writer.WriteValue(this);
+    }
+
+    void Handle(in NetworkContext context)
+    {
+        // Handler code
+    }
+
+    [MessageHandler(MessageType.Custom + (int)CustomMessageType.MyMessage1)]
+    static void Receive(ref FastBufferReader reader, in NetworkContext context)
```
SamuelBellomo, line 457: I feel like having the Receive handling directly in the message would make it a bit more cumbersome to have multiple handlers for that message? Having a subscribable Action might be a good addition?

text/0026-serialization.md
```diff
@@ -0,0 +1,598 @@
...
+}
+
+struct Message : IMessage
+{
+	MessageType GetMessageType()
+	{
+		return MessageType.Custom + (int)CustomMessageType.MyMessage1;
+	}
+
+    int GetSizeUpperBound()
```
SamuelBellomo, line 441: I’m also wondering if some of the boilerplate could be abstracted away, like GetSizeUpperBound() and GetMessageType() . I get that you can’t have a base class with structs. I’m wondering if that could be calculated for you instead.
As a user writing custom messages, I wouldn't really care about these, I'd just care about my message's content and how I handle it on the receiving end.

text/0026-serialization.md
ShadauxCat: Receive doesn't have to be directly in the message, but there is only one handler. But, for clarity, `Receive` is responsible for deserializing the message and then passing it to the handler. I don't think it makes sense for some messages to have multiple handlers, but where it does make sense, `Receive` is where calling those subscribed actions should happen.

The reason for this is to avoid a coupling between the transport-level networking and the message handling code. By having `Receive` be a generic function that takes just the buffer, we can avoid big switch statements that have to know about every type of message, and that's what makes it possible for end users to directly utilize the messaging system. But because this function is where the deserialization happens, I don't think there should be multiple subscribers at this layer - otherwise the reader has to reset its position in between each handler and each handler has to repeat the serialization.

The implementation of the message could be something like this (and this is actually what we did in our hackweek project that was a proof of concept for this design):

```C#
    public struct DummyMessage: IMessage
    {
        public delegate void OnReceivedDelegate(int from, in DummyMessage message);
        public static OnReceivedDelegate OnReceived;
        
        public int foo;
        public int bar;
        
        public void Serialize(ref FastNetworkWriter writer)
        {
            writer.WriteValue(this);
        }

        public MessageType GetCommandType()
        {
            return MessageType.Dummy;
        }
        
        [MessageHandler(MessageType.Dummy)]
        public static void Receive(int from, ref FastNetworkReader reader)
        {
            reader.ReadValue(out DummyMessage command);
            OnReceived?.Invoke(from, command);
        }
    }
```

(btw, this part of the design is moving to a different RFC/different PR - removed from this one, but I'm still pushing for it)

text/0026-serialization.md
ShadauxCat: Not sure how to abstract either of those away.

`GetMessageType()` should probably be a property and not a method, but each and every message struct has to have a different return value from that because that's what goes in the message header and how the handler is looked up on the receiving end.

`GetSizeUpperBound()`... there's no generic implementation for this that would work. The size has to be known before serializing to a buffer, but imagine a message like this:

```C#
struct Message : IMessage
{
    bool HasValue;
    SomeBigStruct Value;
    List<int> VariadicData;
    
    void Serialize(ref FastBufferWriter writer)
    {
        writer.WriteValue(HasValue);
        if(HasValue)
        {
            writer.WriteValue(Value);
        }
        writer.WriteValue(VariadicData.Count);
        foreach(int dataItem in VariadicData)
        {
            writer.WriteValue(dataItem);
        }
   }
}
```

The size of that message is `sizeof(bool) + (HasValue ? sizeof(SomeBigStruct) : 0) + sizeof(int) + sizeof(int) * VariadicData.Count`. I don't see any way that could be guessed in a generic way.

We could avoid the need for `CalculateSizeUpperBound()` by serializing to a temporary buffer and copying after all the data's been written, but that's not really ideal performance-wise...

text/0026-serialization.md
```diff
@@ -0,0 +1,242 @@
...
+```
+
+```C#
+void Serialize(ref FastBufferWriter writer)
+{
+    writer.WriteValue(this);
+}
+```
+
+The former will create more efficiently packed data in the message, and can be even further optimized by using `WriteSinglePacked()` and `WriteInt32Packed()`, but it has two downsides: first, that it involves more method calls and more instructions, making it slower; and second, that it creates a greater opportunity for the serialize and deserialize code to become misaligned, since they must contain the same operations in the same order.
```
MFatihMAR, line 64: I might agree with the first part (more instructions and slower) (even though just saying slower doesn't make it undesirable immediately — how much slower? would be a better question to start thinking about better packing for bandwith vs less CPU cycles trade-off), but on the second point, I'm not sure if I understand the reason behind that argument. `Serialize()` is the method that both serializes and deserializes, there is no 2 different methods, so how could they get out of order / misordered?

text/0026-serialization.md
```diff
@@ -0,0 +1,242 @@
...
+    void Serialize(ref FastBufferWriter writer)
+    {
+        writer.WriteValue(embedded);
+        writer.WriteSinglePacked(f);
+        writer.WriteInt32Packed(i);
+    }
+}
+```
+
+This allows the four bytes of the embedded struct to be rapidly serialized as a single action, then adds the compacted data at the end, resulting in better bandwidth usage than serializing the whole struct as-is, but better performance than serializing it one byte at a time.
```
MFatihMAR, line 106: +1 I'd agree with offering both options and leaving it up to dev to decide which one suits them best in their specific context/use-case.

text/0026-serialization.md
```diff
@@ -0,0 +1,242 @@
...
+# Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+## FastBufferWriter and FastBufferReader
+
+FastBufferWriter and FastBufferReader are replacements for the current NetworkWriter and NetworkReader. They have much the same interface, but there are some critical differences:
+
+- FastBufferWriter and FastBufferReader are **structs**, not **classes**. This means they can be constructed and destructed without allocations if necessary.
+- FastBufferWriter and FastBufferReader both wrap `NativeArray<byte>`, allowing us to take advantage of extremely fast `Allocator.Temp` and `Allocator.TempJob` allocations to create temporary buffers.
```
MFatihMAR, line 117: just to clarify, `FastBufferWriter`/`FastBufferReader` wouldn't create `NativeArray<byte>` instance, it'd be passed into them — so, allocator would be determined from outside, right? (such as having persistent `NativeArray<byte>`s and recycling them at different frames etc.)

text/0026-serialization.md
```diff
@@ -0,0 +1,242 @@
...
+- Neither FastBufferReader nor FastBufferWriter inherits from nor contains a `Stream`. Any code that currently operates on streams will have to be rewritten.
+- FastBufferReader and FastBufferWriter are heavily optimized for speed, using aggressive inlining and unsafe code to achieve the fastest possible buffer storage and retrieval.
+- FastBufferReader and FastBufferWriter use unsafe typecasts and `UnsafeUtility.MemCpy` operations on `byte*` values, achieving native memory copy performance with no need to iterate or do bitwise shifts and masks.
+- FastBufferReader and FastBufferWriter are intended to make data easier to debug - one such thing to support will be a `#define MLAPI_FAST_BUFFER_UNPACK_ALL` that will disable all packing operations to make the buffers for messages that use them easier to read.
```
MFatihMAR, line 121: <3 this!

text/0026-serialization.md
```diff
@@ -0,0 +1,242 @@
...
+
+`BufferSerializer` has an additional advantage (or disadvantage, depending on your viewpoint) compared to `FastBufferWriter` and `FastBufferReader`: since it's intended to be used by our end users, it performs bounds checking on every operation and throws exceptions if attempting to read or write out of bounds. This, combined with the `IsReader` if-checks, will make `BufferSerializer` slower than `FastBufferReader` and `FastBufferWriter` - but for advanced users, we can expose `FastBufferReader` and `FastBufferWriter` and allow the user to use them directly if they choose (with all appropriate warnings about them delivered).
```
MFatihMAR, line 152: I think we can still make `BufferSerializer` to not check boundaries on release builds but do checks on Editor & Development builds. I also believe we should expose `FastBufferReader`/`FastBufferWriter` to public too.

(Why we call them `FastXXX` and not just simply `BufferReader`/`BufferWriter` directly?)

text/0026-serialization.md
```diff
@@ -0,0 +1,242 @@
...
+- The proposed design uses fixed-size buffers and requires manual bounds checking, which may make it a little more difficult to use.
+
+However...
+
+# Rationale and alternatives
+
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- This approach achieves extremely fast performance
+- This approach avoids memory allocations at the core layer except by way of `NativeArray`'s `Allocator.Temp` and `Allocator.TempJob`, which are orders of magnitude faster.
```
MFatihMAR, line 170: I'd repeat myself here. I think reader/writer don't have to allocate underlying NativeArray but just simply use what's being passed into them. They can operate on whatever they're handed and that'd give us the ultimate power of managing memory however we want based on the context (job, frame etc.).

text/0026-serialization.md
SamuelBellomo: For GetMesssageType, you could use reflection and use the message class itself as an identifier. If I recall correctly, you can get either a string for the full class name or a hash for that Type. Then order all your Types by hash and increase your counter to generate a shorter value to send. Since you'll have the same message types both sides, that number should be the same on both sides?
For GetSize, we need the size before we serialize? Else couldn't we get that info from the result of Serialize?

text/0026-serialization.md
ShadauxCat: Serialize doesn't both read and write. In the IMessage proposal, Serialize writes, the MessageHandler reads. And the reason for that is that, in order to decouple message deserialization from incoming data processing (which I propose both because it simplifies maintenance by not having a massive switch statement to maintain, and because it allows us to extend IMessage functionality to user code as a better way to do custom messages), the data processing code has to not know which type it's deserializing... Ergo it can't call any instance methods. 

Also having Serialize go both ways introduces performance overhead by requiring a branch on "IsReading" in every operation... I'm trying to avoid said branches. 

As far as performance cost, it's not very high given all the methods are inlined, and I've also improved the speed of our int packing algorithm a bit. But it's not zero. I measured it a while back, but I've made some improvements for both methods since then, so I'll get some more up-to-date numbers for you.

text/0026-serialization.md
ShadauxCat: They have two constructors. One takes an existing NativeArray, the other takes a size and an allocator. In both cases you get to pick the allocation type.

ShadauxCat: Well... Writer has two constructors... Reader has just the one.

text/0026-serialization.md
ShadauxCat: I don't think it'd be safe to have `BufferSerializer` completely skip bounds checking in release. 

Imagine this scenario: a string is serialized as "length | contents". Someone serialized 999999999999 | "". Reading this without bounds checking is a crash in the best case, and in the worst case, we just added heartbleed to MLAPI. Bounds checking on the receiving side has to be done... For FBW/FBR it's done in batches for performance. For `BufferSerializer`, since that's aimed at less experienced users, I wanted to make the interface more usable even though it'll be slower.

text/0026-serialization.md
ShadauxCat: Yes. You're correct here. Writer's second constructor is just for convenience.

text/0026-serialization.md
ShadauxCat: I personally don't want to go down the road of using reflection / type hashes. There's too many ways that kind of magic can go wrong in my experience. When we start doing type hashes, there's too many ways a code change could break that and now the server can't ever support older clients. I think it's better to be explicit about what value you want in your header.

GetSize... the reason for that was message batching. I wanted to know the size before serializing so I could know if it'll fit into the current message batch buffer or if I needed to make a new one. I wanted to do that without incurring the cost of an extra memcpy by serializing to a temporary buffer and then copying it... but ultimately when I updated the IMessage RFC I decided that simplifying the interface was probably more valuable than avoiding that extra copy, so I removed `CalculateSizeUpperBound()`

Can we continue this discussion on that RFC if you have more feedback?

#29

ShadauxCat: (Also please re-read it, I made some other changes as well.)

text/0026-serialization.md
ShadauxCat: As far as "why not just Buffer reader / BufferWriter"... Why not just "car" instead of "Ferrari"? ;)

Serious answer... They were originally called FastNetworkReader and FastNetworkWriter because 1) I wanted them to be able to sit alongside the existing classes while in development, and 2) I wanted user code using the existing classes to break with "class no longer exists" rather than "class is now a strict and you're not passing with ref so it suddenly breaks at runtime and you don't understand why." When the request came to name it generically, the Fast just stuck. 

But generally speaking, when the main feature of something is its speed, I don't think it's bad to advertise "this is why you might want to use this instead of BufferSerializer, but also why the api is a little non-standard and some features aren't supported" in the name. Makes the design goal clear.

text/0026-serialization.md
ShadauxCat: Ok, I have numbers for you and they widely differ between different kinds of builds.

I did these tests on a small struct:

ShadauxCat: Tested again with something double this size. Results:

Editor:

- Test 1 took almost exactly the same amount of time it took with 4 values
- Test 2 took 210% more time vs Test 1
- Test 3 took 609% more time vs Test 1

text/0026-serialization.md
```diff
@@ -0,0 +1,242 @@
...
+      writer.WriteByte((byte)(value >> 56));
+  }
+  ```
+  
+  Each of those calls to `WriteByte()` will have an additional call to `EnsureCapacity()`, which verifies there's enough space and resizes and copies the array if necessary, resulting in significant method call overhead and possible allocations. This is taken even further than UNet's implementation - where UNet would check for 8 bytes of space in a single check during `WriteByte8()`, Mirror checks for space before writing each and every individual byte. The same comments apply here as the ones I made discussing UNet.
+  
+
+# Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
```
mattwalsh-unity, line 234: The one thing I didn't see above but don't want to hold up the RFC for is mention of code generation.  Can I assume  there's nothing in the approach that would stand in the way of a code generation approach later?

mattwalsh-unity (approved):

text/0026-serialization.md
ShadauxCat: Also worth noting that the int values I was writing for `a` were small (just 127), so in the packed tests, they weren't even hitting the slowest path for int packing. When they do hit the slowest path, things get even worse. Setting the int and float values all to 12345678 in the double size structs I used for my last tests results in Test 3 taking **929%** more time vs Test 1.

MFatihMAR: so, calling write multiple times is not too bad but XXXPacked() calls are 3x-9x more expensive compare to non-packed ones, right? that's not great :/

I think it's fair to comment on doing both read & write in .Serialize() wrt branching. There's a convenience vs performance dilemma there for sure. I can't really pick whether we should separate read/write or not but my gut tells me that we kind of already have/need to have both. BufferSerializer (or NetworkSerializer) might offer you .Serialize() doing both read/write based on its mode (isReading check). On the other hand BufferReader/BufferWriter could offer you a split control for serialization with their NativeArray accepting constructors operating on that array to perform read/write without isReading check. Would that be a fair design you would think?

text/0026-serialization.md
MFatihMAR: yeah, after a quick thought, I also realized no boundary checks means "buffer exploit me please!" by design.

Our current names are `NetworkBuffer`/`NetworkSerializer`/`NetworkReader`/`NetworkWriter` and we might simply re-use them. User code for `Buffer[Serializer/Reader/Writer]` simply does not exist today, we might use that convention too. We tend to pick `NetworkXXX` prefix for things all over the place but this case might be an exception, or not. I'm not that picky about names all the times. We could figure that out on the PR side too. No big deal.

text/0026-serialization.md
MFatihMAR: _(FWIW, I also do not like type-based maps/hashes too)_

text/0026-serialization.md
```diff
@@ -0,0 +1,242 @@
...
+	bitWriter.WriteBits(b, 5);
+} // Dispose automatically adds 2 more 0 bits to pad to the next byte.
+```
+
+## INetworkSerializable
+
+A new `BufferSerializer` class wrapping `FastBufferReader` and `FastBufferWriter` will replace the existing `NetworkSerializer` to provide the same two-way interface as the existing `NetworkSerializer`, and is rewritten as a **ref struct** so that no allocation overhead is incurred in calling the `INetworkSerializable`'s `Serialize()` method. A **ref struct** is chosen instead of a normal struct to avoid the pitfalls associated with having mutable structs and forgetting to pass them with the ref keyword - we can manage this with `FastBufferReader` and `FastBufferWriter` because those are internal and we can rely on our ability to understand the implications of using them as structs, but our end users may be less experienced and may inadvertently pass this forward to calls that do not take the value by ref. A `ref struct` is always passed by reference rather than copied, but does have the disadvantage of being limited to the stack frame - so we can't use it for `FastBufferWriter` and `FastBufferReader` because we persist those.
```
MFatihMAR, line 150: I'm sorry to bother you with this (because I could do a local repro to understand this better).

Are we saying we'd accept `ref struct` only to preserve ability to mutate the source struct being passed into our methods? Is that a requirement for our readers because we do not want to alloc for types but rather write "into" the instance user passed into our reader? What if we were to split `.Serialize()` into two functions here in this case and call them `Serialize(with-no-ref-struct-param)` & `Deserialize(with-ref-struct-param)` to cover both cases?

Well, if I'm misunderstanding this thing entirely, I think a few small code snippets in this section educating me about the issue would be super helpful :)

text/0026-serialization.md
ShadauxCat: That is the exact design I proposed. :) BufferReader / BufferWriter are faster but harder to use, BufferSerializer is slower but more convenient and wraps Reader/Writer. My proposal is we use the faster ones internally, give our end users the convenient one, but also give them the power to pull the underlying Reader/Writer out of it if they want/need extra speed.

text/0026-serialization.md
ShadauxCat: I don't want to reuse the name because I think the difference will confuse people. I was also asked not to use the word "network" because there's a possibility we might split it out in the future to its own package so other teams can use it. I'm ok with dropping Fast if you don't like it, but propose keeping Buffer instead of Network.

text/0026-serialization.md
ShadauxCat: I personally am not a big fan of code gen for serialization... But that's just because I'm a control freak and don't like black boxes or things I can't tweak and optimize to the tiniest detail. But that's just my personal preference, and no, there's nothing that would stop us implementing it with this.

text/0026-serialization.md
MFatihMAR: oh, alright. I didn't know about non-network prefix plans for those. I think I'd prefer Buffer with no Fast in it since we never used those names so far already — but I'd leave it up to you for the final call.

text/0026-serialization.md
ShadauxCat: The reason to make it a ref struct is that the other ones (BufferReader and BufferWriter) being regular structs makes them landmines for inexperienced programmers since they have to be passed by ref always or things break (i.e., passing without the ref keyword creates a copy, which means you end up with two writers or readers pointing at the same buffer but tracking position separately and your writes get lost or overwritten). To me that's an acceptable thing for us to take on internally but not great for us to pass to our users. The suggestion to make BufferSerializer a ref struct is to disarm that landmine - ref strict S still don't have to be allocated (they're stack memory/value types), but they're passed by ref implicitly like classes are, so users can't screw up by forgetting the ref keyword. 

The downside, and reason we can't use them for Reader and Writer, is that ref structs can't outlive the stack context where they were created - they can't be boxed, can't be fields in a class or struct, can't be returned from the function, etc. When the function returns they are destroyed and that's that. But given BufferSerializer is just a wrapper class with helper methods, it doesn't need any persistence, so those limitations don't hurt it and they keep users from being able to make mistakes with it.

text/0026-serialization.md
ShadauxCat: I personally like keeping Fast only because it serves as a contract to the user and a design mandate. When someone goes to add a method to it, they have to think "is this way fast or is there a better approach?" But I'm really not stuck on it. Which is unfortunate because we're in a live lock now, both of us politely saying "You go ahead." "No, please, after you." "No no, I insist, you first."

Someone needs to break our tie here and tell us which of us has to accept our own opinion. :P

text/0026-serialization.md
MFatihMAR: I see, I think we can live with ref struct BufferSerializer, it makes sense. is there a C# doc that I could read more about this stuff? 👀

MFatihMAR: huh, I got this: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/struct#ref-struct

text/0026-serialization.md
MFatihMAR: I accept with your opinion, it is `FastBufferReader` & `FastBufferWriter`.

MFatihMAR (approved):
Looks good! 🚀
(and thanks for clarifications in the PR comments).

The only minor nit-pick I have left is RFC name, I'd call it "Buffer Reader, Writer & Serializer" (+`0026-buffer-read-write-ser.md` I guess) instead of generic "Serialization".

NoelStephensUnity (approved):
Looks great!

LukeStampfli (approved):
Looks all good, I agree with the proposed design.

PR #27, opened 2021-07-23
Rfc/snapshot system

jeffreyrainy:
Snapshot System, spawn part


text/0000-snapshot-system.md
```diff
@@ -0,0 +1,207 @@
...
+- NetworkVariables values at specific ticks
+- Spawn commands, with all the spawn parameters, the tick and a bitmask of target connections 
+
+![](0000-snapshot-system/snapshot.rfc.png)
+
+### The Snapshot Message 
+
+Machines will exchange Snapshot Messages. A Snapshot Message is nothing but a partial Snapshot (plus a header). Those messages will be used to synchronize the state and replicate events. Apart from containing less elements than Snapshots, the only difference between Snapshot Messages and Snapshot is that Snapshot messages don't need the bit mask for destination.
+
+Additionally, a Snapshot Message will include a header that contains:
```
LukeStampfli, line 48: Long term we will need to change that format into something which uses much less bandwidth. But we can address this as a separate feature later.

text/0000-snapshot-system.md
```diff
@@ -0,0 +1,207 @@
...
+- a sequence number unique to this message
+- the tick it was sent on
+- acknowledgment for previous messages from this connection (in the form of a bitmask over the _n_ previous sequence numbers)
+
+![](0000-snapshot-system/snapshot.message.rfc.png)
+
+### Eventual Consistency
+
+The model we will use for synchronisation is eventual consistency. Any Snapshot Message might be lost, but retransmits(1) will have the information eventually arrive to its destination. In the case of packet loss, some variables still get to update.
```
LukeStampfli, line 57: I'm still not sold on this but fine with going forward. I think full state /delta snapshots are better for co-op and other small scale games.

text/0000-snapshot-system.md
```diff
@@ -0,0 +1,207 @@
...
+The first one will control whether NetworkVariables are sent via the Snapshot System or not, and the second one whether the Spawns are sent via the Snapshot System.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+### Guiding principles
+
+The first guiding principle to introduce is that under normal operations **no** memory allocations are done at runtime, within the Snapshot System. Every buffer is pre-allocated when the Snapshot System is started. Provisioning is made for sending, for receiving, and for any internal processing.  
+
+Another guiding principle is that the Snapshot System doesn't care whether we have a star topology, a full mesh, or anything in between. Rather it is only concerned about:
```
LukeStampfli, line 88: I think from a design perspective this is kind of correct. But in practice we would need to add a lot of additional logic to support multiple writer issues in a mesh topology.

text/0000-snapshot-system.md
```diff
@@ -0,0 +1,207 @@
...
+
+In the example above, we'll assume two clients, Alice and Bob, connected to a host, Charlie.
+
+In order to perform its job, the Snapshot System will maintain the following data: (the usage of this data will be presented later)
+
+On hosts, only, for each connection (client):
+- List of interesting objects (AoI-provided)
+
+On all machines, for each connection:
+- For each object: acknowledged spawned status, with tick.
```
LukeStampfli, line 109: Does that include just spawned objects on clients? Or will a client have track over all existing objects in total. Like In a battle royale will the client keep a list of every chest in the entire world?

LukeStampfli (approved):
We already had a few conversations about this in the snapshot sub team and this RFC reflects what we discussed and I agree with the design.

mattwalsh-unity (approved):

PR #28, opened 2021-07-25
NetworkTransform Interpolation

SamuelBellomo:
See this [page](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/network-transform-smoothing%2Fbuffered-interpolation/text/0000-network-transform-buffered-interpolation.md) for formatted view


text/0000-network-transform-buffered-interpolation.md
```diff
@@ -0,0 +1,408 @@
...
+```C#
+        public AnimationCurve DistanceSendrate = AnimationCurve.Constant(0, 500, 20);
+```
+configuration (that was used to set interpolation times) to use instead a swappable interpolator.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+NetworkTransform now has this new configuration:
+- `Interpolator`: Allows users to select the interpolation algorithm they want to use. This interpolator will interpolate "ghost" side state changes. Note it'll still be up to users to interpolate their authoritative objects however they want. If for example Client 1 is authoritative over a player object, this RFC will add interpolation to the server object and the other clients' objects, not Client 1's object. It'll still be up to users to interpolate that authoritative object (using RigidBody's interpolation for example or anything else).
```
LukeStampfli, line 42: This does not sound like great UX to me. Since the idea is that the position changes to the NetworkTransform happen during a fixed step (be it network tick or FixedUpdate doesn't matter) I think we should consider providing a way to auto interpolate the authoritative objects as well by using the same interpolation as a RigidBody.

text/0000-network-transform-buffered-interpolation.md
```diff
@@ -0,0 +1,408 @@
...
+An `IInterpolator` interface has been added, to allow users to create their own custom interpolator.
+The implemented interpolator needs to keep its own state and will be given new values each time ones are available to the NetworkTransform's OnValueChanged.
+
+```C#
+public interface IInterpolator<T>
+{
+    public void Update(float deltaTime);
+    public void NetworkTickUpdate(float tickDeltaTime);
+    public void AddMeasurement(T newMeasurement, NetworkTime sentTime);
+    public void Teleport(T value, NetworkTime sentTime);
```
LukeStampfli, line 71: Teleport sounds very position oriented. How about `Reset`?

text/0000-network-transform-buffered-interpolation.md
```diff
@@ -0,0 +1,408 @@
...
+
+https://user-images.githubusercontent.com/71790295/126731436-f9e89d75-4973-49e8-bbb5-84ff67c48a9b.mov
+ 
+An `IInterpolator` interface has been added, to allow users to create their own custom interpolator.
+The implemented interpolator needs to keep its own state and will be given new values each time ones are available to the NetworkTransform's OnValueChanged.
+
+```C#
+public interface IInterpolator<T>
+{
+    public void Update(float deltaTime);
```
LukeStampfli, line 68: Is there a reason why need to make the interpolator stateful here? Or could we also make this work by passing in an absolute time value?

LukeStampfli: Also trying to figure out in what situation you would need `Update` + multiple `GetInterpolatedValues` Maybe just a single `GetInterpolatedValue` function which takes an absolute time value would be a better approach?

text/0000-network-transform-buffered-interpolation.md
```diff
@@ -0,0 +1,408 @@
...
+
+You can find the interpolator's configuration in its associated scriptable object factory.
+It is advised to use the same interpolator for all your NetworkTransforms. This way, you're making sure all your objects will be synchronized and use the same delays. Having an interpolator with 100ms delay and another with 500ms delay could cause visible overlaps and desyncs where an object at time say `t=10s` tries to interact with an object buffered at time `t=9.6s`.
+
+If you have a case where you'd like to use a different configuration, you can create a new instance of the interpolator scriptable object and assign it to your object. If you want a new default value for all your NetworkTransforms, you can assign a scriptable object as default value for a class in the Unity Editor, by selecting the C# script and assigning the default value in the script's inspector (this will add a reference in the C# script's meta file).
+
+If you don't want interpolation, you can use the `NoInterpolation` interpolator which will take in new values and present them directly when asked, without doing anything else.
```
LukeStampfli, line 54: I assume leaving the factory serialize field empty on NetworkTransform will default to this `NoInterpolation`?

text/0000-network-transform-buffered-interpolation.md
```diff
@@ -0,0 +1,408 @@
...
+
+## Prediction and interpolator.
+Current interpolator only affects ghosts and doesn't affect locally owned objects. An eventual predicted player wouldn't be affected by this interpolation.
+Future prediction might want to interpolate corrected values instead of teleporting players. This should be handled by prediction code.
+Since interpolator is a self contained class and isn't a MonoBehaviour, interpolators could be reused for that correction interpolation.
+
+## Other APIs for interpolator
+The Interpolator's `Update()` method could have been injected with the last 3 positions. 
+```C#
+Vector3 Update(Vector3 pos1, Vector3 pos2, Vector3 pos3)
```
LukeStampfli, line 275: I prefer the current API, It allows to implement an interpolator which uses any number of data points. You could do 3 positions with the current API.

LukeStampfli (commented):
I think I can agree with the overall idea behind this.

What is a bit worrying is that the Interpolator does a lot of things in addition to interpolating. It is also storing a state history and is aware of frames/ticks. I can see why we are doing it to quickly fix issues in the NetworkTransform but I think we will need get back to this and redesign it when we get snapshots.

text/0000-network-transform-buffered-interpolation.md
jeffreyrainy: I would like to see more detail of when each function is called. In addition to Luke's point of removing update and having `GetInterpolatedValue` take the time to interpolate at, I'd like to see:

`AddMeasurement` is called when... 
`Teleport` is called when...

and in fact, why do we need `NetworkTickUpdate`? If `GetInterpolatedValue` takes the time, it might not even be needed and make the interpolator agnostic of ticks.

text/0000-network-transform-buffered-interpolation.md
SamuelBellomo: you would still need different interpolation algorithm for each. Buffered interpolation adds artificial latency. if I have 100ms of buffer, that'd mean the host would have 100ms of latency on its authoritative object for no reason.
There's not really any reason for NetworkTransform to touch an authoritative object. NetworkTransform is to ghost game owned objects.
The lerp a user would have needed to add in a singleplayer game would still be needed here. They'd just add NetworkTransform to "just make it network"

text/0000-network-transform-buffered-interpolation.md
SamuelBellomo: > Also trying to figure out in what situation you would need Update + multiple GetInterpolatedValues Maybe just a single GetInterpolatedValue function which takes an absolute time value would be a better approach?

ApplyNetworkState (which calls GetInterpolatedValue) is called two times right now. Once on Update like you suggest and once when reseting any ghost side changes a user might have done. Having this Get without doing any changes to the stateful interpolator was useful when I tested Kalman Filtering extrapolation. I could call Update once and then get the resulting extrapolated value multiple times without fearing changing my interpolator state.

I added details on when Get is called
I did the same with Add and Reset

> and in fact, why do we need NetworkTickUpdate? If GetInterpolatedValue takes the time, it might not even be needed and make the interpolator agnostic of ticks.
So for me, this is about flexibility for whoever wants to implement their own interpolator. As you can see in the list of possibilities, there's a few other options. I could see a need for a physics based interpolator needing a FixedUpdate call. (looking at my current interface, looks like I'm missing that one, I'll add it in addition to the network update). NetworkTickUpdate isn't currently used my provided interpolators (NoInterpolator and BufferedInterpolator), but I can see new user interpolators using it.

text/0000-network-transform-buffered-interpolation.md
SamuelBellomo: I would do smooth by default, so I'd use BufferedInterpolator by default.
I'll check with UX for a final approval, but on your side, what do you think?

SamuelBellomo:
> I think I can agree with the overall idea behind this.
> 
> What is a bit worrying is that the Interpolator does a lot of things in addition to interpolating. It is also storing a state history and is aware of frames/ticks. I can see why we are doing it to quickly fix issues in the NetworkTransform but I think we will need get back to this and redesign it when we get snapshots.

It's very possible we'll need to break the API with snapshot, but I'm not sure. Passing a Vector3 for position to AddMeasurement would still happen, snapshot or no snapshot. Passing a NetworkTime would still happen snapshot or no snapshot.
I would imagine the snapshot system would still be available statically from the interpolator, without needing to inject snapshots through the interface. If there's any need for rewinding things, interpolators could still have access to that information.
Our usage of the interpolator might change, but I think the interface is flexible enough to give us some leeway once we add snapshots later? Unless you have some other ideas of things we could work into the interface?

jeffreyrainy:
> > I think I can agree with the overall idea behind this.
> > What is a bit worrying is that the Interpolator does a lot of things in addition to interpolating. It is also storing a state history and is aware of frames/ticks. I can see why we are doing it to quickly fix issues in the NetworkTransform but I think we will need get back to this and redesign it when we get snapshots.
> 
> It's very possible we'll need to break the API with snapshot, but I'm not sure. Passing a Vector3 for position to AddMeasurement would still happen, snapshot or no snapshot. Passing a NetworkTime would still happen snapshot or no snapshot.
> I would imagine the snapshot system would still be available statically from the interpolator, without needing to inject snapshots through the interface. If there's any need for rewinding things, interpolators could still have access to that information.
> Our usage of the interpolator might change, but I think the interface is flexible enough to give us some leeway once we add snapshots later? Unless you have some other ideas of things we could work into the interface?

What I hope happens is, once Snapshot is there and works, for the interpolator to attach to the NetworkVariable directly. Then the NetworkTransform responsibility becomes ones of integrating with the Unity's transforms and integration with UI, surfacing parameters type stuff. But I hope the interpolation to eventually go down into NetworkVariables. Would still use the/an interpolator, but just not tied to NetworkTransform, if that makes sense.

SamuelBellomo:
> > > I think I can agree with the overall idea behind this.
> > > What is a bit worrying is that the Interpolator does a lot of things in addition to interpolating. It is also storing a state history and is aware of frames/ticks. I can see why we are doing it to quickly fix issues in the NetworkTransform but I think we will need get back to this and redesign it when we get snapshots.
> > 
> > 
> > It's very possible we'll need to break the API with snapshot, but I'm not sure. Passing a Vector3 for position to AddMeasurement would still happen, snapshot or no snapshot. Passing a NetworkTime would still happen snapshot or no snapshot.
> > I would imagine the snapshot system would still be available statically from the interpolator, without needing to inject snapshots through the interface. If there's any need for rewinding things, interpolators could still have access to that information.
> > Our usage of the interpolator might change, but I think the interface is flexible enough to give us some leeway once we add snapshots later? Unless you have some other ideas of things we could work into the interface?
> 
> What I hope happens is, once Snapshot is there and works, for the interpolator to attach to the NetworkVariable directly. Then the NetworkTransform responsibility becomes ones of integrating with the Unity's transforms and integration with UI, surfacing parameters type stuff. But I hope the interpolation to eventually go down into NetworkVariables. Would still use the/an interpolator, but just not tied to NetworkTransform, if that makes sense.

I'm still not convinced about moving this in NetVar. I had to move transform state sending to NetworkTick in NetworkTransform in order to make this fit with BufferedInterpolation. I'm pretty sure we can't impose this on NetVar update call sites.
Do you have use cases for interpolation at the NetVar level beside transform related ones?

text/0000-network-transform-buffered-interpolation.md
SamuelBellomo: I'd even go as far as adding more events like Awake, Start, OnNetworkSpawn
I wanted to test something with my buffered interpolator poc and I realized I needed OnNetworkSpawn for that. Having all these available, even if unused by our implementation would help be more flexible for our users.

text/0000-network-transform-buffered-interpolation.md
```diff
@@ -0,0 +1,447 @@
...
+
+![](0000-network-transform-buffered-interpolation/jitter.jpg)
+
+This RFC suggests an extensible interface for users to use and create different interpolation algorithms. It also suggests a default buffered interpolator that'll resist to jitter. 
+After this, users should be able to select the interpolation algorithm that best fits the GameObject they want to sync.
+
+In addition, this RFC suggests removing the 
+```C#
+        public AnimationCurve DistanceSendrate = AnimationCurve.Constant(0, 500, 20);
+```
```
MFatihMAR, line 35: nit
```suggestion
```cs
public AnimationCurve DistanceSendrate = AnimationCurve.Constant(0, 500, 20);
```
```

File: text/0000-network-transform-buffered-interpolation.md [hunk 7]
```diff
@@ -0,0 +1,447 @@
...
+
+
+A default `BufferedLinearInterpolator` is provided. The buffered linear interpolator will buffer values before making them available to NetworkTransform's value update. This will allow NetworkTransform to accumulate jittered network values without affecting the transform's state and latter be able to consume them at regular intervals.
+
+This will smooth tranforms on jittery connections, however it'll add more latency to your transform updates. This should be used carefully as to not add more latency than needed to your game. Bigger buffer values could be expected when targeting mobile platforms for example, while keeping interpolation times lower for platforms with typically more stable connections.
+
+You can find the interpolator's configuration in its associated scriptable object factory.
+It is advised to use the same interpolator for all your NetworkTransforms. This way, you're making sure all your objects will be synchronized and use the same delays. Having an interpolator with 100ms delay and another with 500ms delay could cause visible overlaps and desyncs where an object at time say `t=10s` tries to interact with an object buffered at time `t=9.6s`.
```
MFatihMAR, line 49: I believe using the same interpolator and its settings is fair for "same type of NetworkTransforms" but I wouldn't advise to use for "all NetworkTransforms". A car would be interpolated differently compared to a character. It'd make sense to use the same interpolator+config for all cars but it wouldn't make sense to share that interpolator+config with a character walking around.

text/0000-network-transform-buffered-interpolation.md
MFatihMAR: FWIW, I think leaving interpolator field empty/null should not default to `BufferedInterpolator` but instead default to latest pos/rot/sca snapping.

text/0000-network-transform-buffered-interpolation.md
SamuelBellomo: Good point, I'll clarify that. The way I wrote made it seem more like an exception, but like you say these could be very common "exceptions".

text/0000-network-transform-buffered-interpolation.md
SamuelBellomo: I'm curious, why?
I'm not strongly in favor of any one, I wanna build a list of pros and cons I can expose to UX.
For me:
Pros:
it's smooth by default
Cons:
It adds latency by default

MFatihMAR: I actually want to move this conversation from here to completely different space :) I'll comment over there at the motivations section soon :P

text/0000-network-transform-buffered-interpolation.md
```diff
@@ -0,0 +1,472 @@
...
+- Start Date: `2021-07-19`
+- RFC PR: [#0000](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/pull/0000)
+- SDK PR: [#0000](https://github.com/Unity-Technologies/com.unity.multiplayer.mlapi/pull/0000)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes upgraded interpolation options from the old NetworkTransform's interpolation that hides jitter and is more flexible for users.
+
+# Motivation
```
MFatihMAR, line 13: after giving another thought to this, I started to wonder whether or not we should offer interface abstraction, scriptable objects and all that fuss around modularity/flexibility. what we really need is a good default implementation with useful properties to configure and we're good to go. if a user wants to change their interpolation algorithm but leave everything else just as it is, I wonder what they'd be trying to achieve. the fact that we're shipping this high-level component open-source gives me even more confidence regarding "one great default implementation, no extreme flexibility". by having one single default interpolation algorithm implementation, we could further optimize this and tune everything around it. offering flexibility for just flexibility's sake doesn't sound like a sensible approach to me here. if there would be a user-need and/or community ask, we might reconsider interfaces, scriptable objects and such. well, even then, we might suggest people to either fork and change it or simply copy-paste our implementation entirely as a boilerplate and build custom solution/algorithm on top of it by modifying only the parts they'd like if we wouldn't be able to cover their cases underneath our implementation.

TL;DR: I do not think we need interfaces, scriptable objects and all this much of a flexibility. we only need a good default implementation with useful config properties exposed. current proposal seems like an overkill to me.

SamuelBellomo: > by having one single default interpolation algorithm implementation, we could further optimize this and tune everything around it.

I agree having everything in the same implementation would help making this more optimizable.
However, would the perf gain be that great? If I have 1000 NetworkTransforms, we'd be saving 1000 Interpolator.Update() calls per frame for example, is that enough to justify losing on that flexibility?
We also have bigger problems. Because we don't have snapshot system that queriable right now, each NetworkTransform needs to keep its own buffer of "snapshots". This means if I have 1000 NetworkTransforms, I have 1000xBuffers anyway. What's one more interpolator object?
We could use struct instead and not use interface, that could help not entrench ourselves and allow for future optims, but again is it worth it?

I'd still argue having interpolator in a self contained class/struct is best. It'd make it very easy to unit test (this is the kind of thing I'll test the crap out of). It'd also make it easy to reuse. My current PoC uses generic types for buffered interpolation. Luke suggested an interesting use case where a user could have one single netvar `t` that controls the movements of all platforms in your game. The self contained interpolator could be used here to interpolate `t`. Having all that buffering code directly in NetworkTransform would have users dig in our implementation for their own buffering.
Just for testability and reusability, I'd argue it's worth it to have it self contained.

On Interface and ScriptableObjects:
Rereading my motivation section, I realize it's a bit short and I should have consolidated the other points I make later in this RFC.
Let's split this in two: scriptable object usage and extensibility through interfaces.


On extensibility with interface: We'll have users creating games at 10 ticks/sec or even lower. linear interpolation isn't great in that case. Gafferongames has a good example gif of this here https://gafferongames.com/post/snapshot_interpolation/#linear-interpolation with the wobbly children cubes. I don't think there's one interpolation to rule them all, having that interface allows to easily plug your own interpolation.
If you look at the prior-art section, you can see there's already quite a few use cases I can see users needing.
Gafferongames talks about wobbly children and to use Hermite interpolation. Source engine uses buffered linear interpolation like I'm suggesting here. HLAPI used physics based extrapolation. Unreal interpolates visuals only and teleports colliders. Jeff mentioned using Kalman filtering for a more realistic smooth extrapolation.
Having that interface offers some sort of contract to guide users in their own interpolator creation.

I'm on the edge for scriptable objects, I'd tend to agree with you, but here's some additional rationale on why it'd be great for users:
One advantage is the ability to create multiple "global" configs.
I can have a "allCarsInterpolation" and a "allCharacters" interpolation, each with their own settings, but still promote consistency between these types of objects (this way I don't have some characters at 100ms buffering and others at 500ms, I'm sure they all have the same buffering setting).
If you had per NetworkTransform buffer settings for example, it'd be too easy to have an object somewhere where you mistakenly put a different value than everything else. You'd then have an easy to miss desync that'd happen once in a while when these objects cross path. Doing this through global SOs (but still having the option to create new configs if necessary) promotes consistency. It's way easier to see mistake character with a car interpolator than a character with an interpolator setting set to 400 instead of 100.

MFatihMAR: > I don't think there's one interpolation to rule them all, having that interface allows to easily plug your own interpolation.
> ...
> Source engine uses buffered linear interpolation like I'm suggesting here. HLAPI used physics based extrapolation. Unreal interpolates visuals only and teleports colliders.

Which persona and project we're talking about here? Are we expecting devs to write multiple different NetworkTransform interpolation algorithms for their projects? I really can't see 3-5 different custom interpolation algorithms being implemented by majority of our users. That minority who'd write their own interpolator for their specific needs will probable throw-away our implementation and implement a custom NetworkTransform with its interpolation code anyway. Even in the same game project, I can't see more than 2-3 different interpolators implemented and that game would be something like an open-world sandbox game where you can drive vehicles, fly jets, walk around and do melee combat. I'd highly argue that those people working in those types of games will throw-away NetworkTransform entirely to roll their own, not just the interpolation algorithm. You already gave great examples for "one interpolator to rule 'em all!": Valve's Source, Epic's Unreal, Unity HLAPI etc. What would you do if that engine/runtime doesn't cover your interpolation needs by their default implementations? You'd throw away and write your own custom interpolation! And in that case, you probably won't keep NetworkTransform around for long. So, at the end of the day, it still feels like we're trying to solve a problem which doesn't exist for majority of the use-cases, needs.

> ... for scriptable objects, ... here's some additional rationale on why it'd be great for users: ...

I'm not against scriptable objects, what I'm suggesting is simply leaving it up to user entirely. They might find SOs useful and instantiate their interpolation configs from SO assets if they want. That'd be a good sample/tutorial for sure but I'm not sure if putting this onto NetworkTransform's / Interpolator's plate is a good idea. Your contract would only care about its interface, not how backing instance is created/serialized etc. Let's leave it up to devs?

SamuelBellomo: Just to confirm, there's two issues:
Having flexibility for multiple interpolation algorithms
Having scriptable object configs for different settings for the same implementation

I see these types of games being built with Unity:
Typical first person movements with characters moving around: Fine with just buffered interpolation implementation. Could use ScriptableObjects configs for different types of characters. 
Extravagant game with non-standard object movements: Will most likely want custom interpolation. Could also use ScriptableObjects configs for different types of objects.

I disagree it's going to be a minority that'll want to do the second case, but it's just an instinct.
I agree for an FPS style game, linear interpolation might be enough for most moving objects. But Unity devs do so much more more than that. Unity is known for being easy to create weird styles of games. Our users love to experiment and create innovative designs, to me it's one of our greatest differentiator from other engines.
I'd be very interested to get some stats on this. It doesn't seem to me like Gafferongames' child cubes example is too far fetched, it seems pretty common. 

> Your contract would only care about its interface, not how backing instance is created/serialized etc. Let's leave it up to devs?

Just so we're clear, the interface doesn't dictate how interpolators are instantiated. NetworkTransform would have one implementation, but if users wanted to use interpolator for custom code, they wouldn't care about scriptable object.
I could also just make "PositionInterpolator" inside NetworkTransfrom virtual and be done with this. I don't care that much about ScriptableObjects. I do care about the Interpolator interface.

MFatihMAR: I'd like to summon @mattwalsh-unity and @andrews-unity here. They'd bring other perspectives to this discussion.

LukeStampfli (approved):
No more scriptable objects yay 😄

jeffreyrainy (approved):

andrews-unity:
Just so I understand we are still purposing the addition of an interface ? And the ability to set per object different interp algorithms? If so I am going to want to understand why we need this. I truely believe 90% of the time most devs won't change it beyond the default one. So to me we should make that the best possible algorithm and say if you want something special there is the code write it yourself. 

We constantly try to provide this plug and play swap this and that model and I just don't see the argument for it. You mentioned several other engine all of which have one path maybe with some nobs and dials and 100s and 100s of multiplayer games have shipped on them. We don't need to solve for everyone we need to solve for the majority.

SamuelBellomo:
> Just so I understand we are still purposing the addition of an interface ? And the ability to set per object different interp algorithms? If so I am going to want to understand why we need this. I truely believe 90% of the time most devs won't change it beyond the default one. So to me we should make that the best possible algorithm and say if you want something special there is the code write it yourself.
> 
> We constantly try to provide this plug and play swap this and that model and I just don't see the argument for it. You mentioned several other engine all of which have one path maybe with some nobs and dials and 100s and 100s of multiplayer games have shipped on them. We don't need to solve for everyone we need to solve for the majority.

@andrews-unity 
I'm still suggesting an interface, but I'm not suggesting the way it's going to be exposed to users in inspector anymore. The interface would be to serve as a contract for users that do want to extend it.
One near-term use case I can see for a new interpolator would be with Luke's network physics work. There's potential for linear interpolation to not be enough and a need for syncing more data (velocity for example). Having that contract would help extending current interpolation for a physics based one for example, but still keep default linear interpolation for when we want to save on bandwidth.
Unreal teleports character capsules but interpolates graphics. It's great for predicted characters, but not that great for a self contained component like we're trying to build here.
Hermite interpolation uses more bandwidth, but has less wobble.
Kalman Filtering is more physics based and less noisy, but can show transform updates that are not true, with an inertia client side that wouldn't be there server side.
Buffered Linear interpolation would still remain the default. It's smooth even with bad jitter and only uses a float for extra bandwidth to track the time the value was updated.

PR #29, opened 2021-08-06
Zero-alloc messaging flow RFC.

ShadauxCat:
[See the linked RFC document →](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/zero-alloc-messaging/text/0029-zero-alloc-messaging-flow.md)

RFC proposing a new way of defining, sending, and processing messages using FastBufferReader and FastBufferWriter and avoiding allocations.


mattwalsh-unity (approved):

text/0029-zero-alloc-messaging-flow.md
```diff
@@ -0,0 +1,390 @@
...
+}
+```
+
+Additionally, each message should have a method associated with it it as follows:
+
+```c#
+[MessageHandler(MessageType.MyMessageType)]
+public static void Receive(ref FastBufferReader reader, NetworkContext context);
+```
+
```
NoelStephensUnity, line 61: I do like the concept of this.  It forces a messaging scheme across the board.

text/0029-zero-alloc-messaging-flow.md
```diff
@@ -0,0 +1,390 @@
...
+- The secondary reason is performance - structs can be created on the stack with no heap allocations, and with FastBufferWriter and FastBufferReader, they can be rapidly serialized as whole units.
+- The third reason is easier debugging - for situations where a lot of packed integers are being used, a preprocessor flag can be used to switch between optimal byte packing and simple whole-struct serialization. At debug time, when there's no dynamic data involved, the buffer in a reader or writer can be simply cast to the desired type by adding a watch of `*(MessageType*)reader.GetUnsafePtrAtCurrentPosition()`
+- The fourth reason is to improve the interface by which messages are sent - instead of needing a Begin...() to add header information and an End...() to perform sending logic, MessageQueueContainer now simply has `public void SendMessage<TMessageType>(in TMessageType message) where TMessageType: IMessage{}`, which can handle both responsibilities, as well as message batching responsibilities, all in one location.
```
NoelStephensUnity, line 94: Consolidating this is nice too.

text/0029-zero-alloc-messaging-flow.md
```diff
@@ -0,0 +1,390 @@
...
+    public short MessageSize;
+}
+```
+
+Since the largest value is 2 bytes in size, the struct should be aligned to a 2 byte alignment, which should make the struct packed and the total header size equal to exactly 4 bytes, which can be sent and received using `WriteValue` and `ReadValue` without any wasted bandwidth.
+
+In addition to the message header, there's also a batch header:
+
+```C#
+internal struct BatchHeader
```
NoelStephensUnity, line 117: Like this as well!

text/0029-zero-alloc-messaging-flow.md
```diff
@@ -0,0 +1,390 @@
...
+    
+    // Returns true if processing should be delayed, false if the specified stage is the current stage
+    // and thus the message should be processed.
+    public bool DelayMessage(DelayStage stage, ref FastBufferReader reader, NetworkContext context);
+}
+```
+
+During each of those three stages, we'll process any delayed queues.
+
+This uses a static queue of (configurable) fixed size per DelayStage value, so no allocations will be incurred in the queueing itself, **however, generally speaking, in the first version, creating delegates will cause an allocation.** There's no real way around that.
```
NoelStephensUnity, line 153: Would it be horrid to just make DelayedNetworkAction an actual method that is always invoked and have that invoke a virtual method like `protected virtual void OnDelayedNetworkAction();` ?

text/0029-zero-alloc-messaging-flow.md
```diff
@@ -0,0 +1,390 @@
...
+
+The first item in the SendQueue is a persistent buffer (allocated with `Allocator.Persistent`) that contains enough space for a non-fragmenting QoS. This buffer is kept around from frame to frame, so that frames where small amounts of data are sent don't even suffer `Allocator.TempJob` allocations.
+
+### Sending to LocalClientId
+
+When the server sends to its own LocalClientId, some small changes are made to the flow: Instead of creating a buffer with `Allocator.Temp`, it creates it with `Allocator.TempJob`. Then, once the buffer has been serialized, instead of copying it to a `SendQueue` buffer, it bypasses the transport by directly inserting the buffer into the Incoming Message Queue. Messages destined for Timing.End will be processed in the same frame; messages destined for Timing.Start will be processed in the next frame, unless added during processing of another message during EarlyUpdate.
```
NoelStephensUnity, line 224: Yay! A way to handle loopback!

text/0029-zero-alloc-messaging-flow.md
```diff
@@ -0,0 +1,390 @@
...
+
+### Channels/QoS
+
+The way channels work is going to be rethought - the `NetworkChannel` value sent to the transport will represent a direct one-to-one mapping between a single `NetworkChannel` and a single QoS value. The current `NetworkChannel` values will be lofted up to a higher level and will be represented as a byte-sized value in the message header. This allows us to be a little more intelligent about how we handle network channels - specifically because it allows us to be able to to tell when two channels are behaviorally identical and treat them the same at the transport level.
+
+Since messages sent with different QoS guarantees have different behavior requirements, the data for different QoS guarantees cannot be stored in the same buffer. To handle this, each client object stores its current QoS on its object next to the buffer. If any message is sent for a QoS other than the one currently stored, the batch is completed and a new batch is started for the new QoS (i.e., a new NativeArray is added to the send queue).
```
NoelStephensUnity, line 234: !ding!

text/0029-zero-alloc-messaging-flow.md
```diff
@@ -0,0 +1,390 @@
...
+
+![Receiving Messages](0029-zero-alloc-messaging-flow/receiving.png)
+
+## Custom Messages
+
+Custom messages as they are, using buffers, no longer need to exist. Since the message handling code is part of the message definition, users have the freedom to define their own messages. Our internal code doesn't need to know anything about the messages themselves in order to send or receive them, so users can create any custom message they want the same way we define our own internal messages, and send them the same way we send ours, and their message handler will be magically invoked when the message is received. The only gotcha is the MessageType enum - since users can't extend the enum itself, we provide a `MessageType.Custom` that's always at the end of the enum, and users can operate as such:
```
NoelStephensUnity, line 271: Users having the ability to create their own message types with associated IMessage implementation is very cool indeed!

NoelStephensUnity (approved):
A big task to get everything converted over indeed, but once done looks like a very good step forward!  Will look forward to the implementation!  :)

PR #33, opened 2021-08-12
network variables / collections changes (#28)

mattwalsh-unity:
[Formatted RFC Draft here](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/netowrk-collections-and-variables/text/0028-netowrk-collections-and-variables.md)

### DIFF
--- a/text/0028-netowrk-collections-and-variables.md
+++ b/text/0028-netowrk-collections-and-variables.md
@@ -80,7 +80,7 @@ Before this change, `NetworkList`, `NetworkSet` and `NetworkDictionary` supporte
 Before this change a user could simply declare a container and add an arbitrary amount of elements.  This creates at least 2 complications; one, it works against the upcoming goals of having MTU-sized snapshot packets.  Second, having network variables be variable length gets in the way of performance-oriented, allocation-free schemes we have planned.  To support this, a user must specify a maxmimum capacity when instantiating a container.
 
 # Drawbacks
-The old method where clients could simply write to a network variable is simpler, especially when trying to employ a client-authoritative scheme. 
+The old method where clients could simply write to a network variable is simpler, especially when trying to employ a client-authoritative scheme.  One way to think of network programming is that each design is a combination of tradeoffs around ease of use, synchronization consistency and performance.  While we seek to maximize all of them, the pre-RFC design maximizes ease of use at the expense of consistency (different clients & the server can have different values at different times with no way to link them) and performance (network variables rely on being ordered & reliable).  The upcoming snapshot system addresses consistency and performance, but at least at this point, it is intractable to do that with the current (extremely easy) interface on NetworkVariables.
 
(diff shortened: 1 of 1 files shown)


text/0028-netowrk-collections-and-variables.md
```diff
@@ -0,0 +1,94 @@
...
+
+
+### New: Read Permissions
+* `Everyone`
+* `ServerOnly`
+
+Going forward, Network Variables will only be (directly) writable by the server.  Previously, a client could change state on the server by writing to its local network variable instance.  As we move towards the snapshot tick-based model this free-form mechanism becomes intractable to support at least in its current form.  With this change, when a client wants to cause a state change it will need to use a different means such as sending an RPC.  
+
+Hence, with this release there are only *Read* permissions.  Either a variable can be read by everyone, or the specific owner of the variable.  The server can always read all variables.
+
```
NoelStephensUnity, line 37: While I think it is pretty clear that the server is the only one with write permissions, is the reason owner clients can't write to a NetworkVariable due to the issue with change in ownership?

text/0028-netowrk-collections-and-variables.md
```diff
@@ -0,0 +1,94 @@
...
+NetworkVariable<TestClass> TheClass = new NetworkVariable<TestClass>(new TestClass());
+TheClass.Value.SomeBool = false;
+TheClass.Value.SomeInt = k_TestUInt;
+```
+
+This code in fact causes nothing to happen on the server as it isn't in fact calling `set`.  Users must remember to call `TheClass.SetDirty(true)`.  
+
+Moreover, the above code doesn't work at all with struct member access.  With structs, the only way to change value is to assign a new copy.
+
+Instead, with this RFC network variables will no longer have a `set` on the `Value` property.  The `get` property will return the value by reference and mark the variable as dirty.  There will be mitigation around avoiding the fact that dirty-ing will happen more often to avoid unnecessary updates.
```
NoelStephensUnity, line 70: Why mark something dirty if you are just reading it?  Is there an underlying logic behind this I am missing?

text/0028-netowrk-collections-and-variables.md
NoelStephensUnity, line 37: While I think it is pretty clear that the server is the only one with write permissions, is the reason owner clients can't write to a NetworkVariable due to the issue with change in ownership?

NoelStephensUnity (approved):
Other than the two minor questions I had and based on your presentation and deeper discussion around permissions.  For the context of what this is for...LGTM!

rivFox:
If Network Variables will only be writable by server and readable by ServerOnly or Everyone, is the reason to use Network Variables with ServerOnly mode? If I understand correctly, the variable will be not synced. 

Second question is why OnlyOwner doesn't exist for read permission?

Clonkex:
The drawbacks section says something about getting the variable marking it as dirty. What? Did I miss something? Why would that ever be the case?

PR #34, opened 2021-08-18
Device Reconnection in Transports

simon-lemay-unity:
[Formatted text available here.](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/device-reconnection/text/0000-device-reconnection.md)

This is a bit of an unusual RFC in that it's mostly concerned with a feature being added to UTP. The technical implementation details will probably be of little interest to most NGO developers. But it does expose new events in the transport interface, which is what we're most interested in in terms of feedback (although we'll be happy to receive comments on the rest).

The RFC proposes to expose `Reconnecting` and `Reconnect` events from the transport. The assumption behind exposing these is that they might be useful to NGO to handle lag spikes and short losses of traffic (whether caused by a device reconnection or not). But if the overwhelming feedback is that NGO will never make use of these events, we might reconsider our decision of exposing them (allowing us to avoid being tied to a particular implementation strategy for device reconnection).


text/0000-device-reconnection.md
```diff
@@ -0,0 +1,162 @@
...
+- Building other protocols on top of the base UDP protocol, which can support seamless mobility. For Relay that would require the server to implement that base UDP protocol, which they might not be able/willing to do. This solution would also prevent users from creating fully encrypted connections (since control messages would be unencrypted).
+
+Not implementing device reconnection means that users will be inconvenienced (by having their online sessions stop working) when roaming from one network to another, which is bad UX. Not providing the feature would also go against our goals of being a top-of-the-class multiplayer solution with a mobile-first philosophy.
+
+# Prior art
+[prior-art]: #prior-art
+
+- The QUIC protocol ([RFC 9600](https://datatracker.ietf.org/doc/html/rfc9000)) is a network protocol built over UDP that automatically handles reconnections the same way we do for the base UDP protocol. (Obviously QUIC is much more complex since it also bundles TLS and congestion control.)
```
nlebedenco, line 148: Typo here. Should read RFC9000 instead of RF9600 the http link is correct though.

simon-lemay-unity: Fixed. Thanks for catching that.

PR #35, opened 2021-08-31
Physics

LukeStampfli:
[See linked RFC document →](https://github.com/Unity-Technologies/com.unity.multiplayer.rfcs/blob/rfc/physics/text/0035-physics.md)

RPC proposing a physics solution for co-op games.


andrews-unity (approved):
Overall this looks fine to me..

jeffreyrainy (approved):
That all seems reasonable to me.

I think we'll need a bit fancier ways to synchronize physics in the future (there bigger fish to fry first, though), but this is a good solid start.

mattwalsh-unity (approved):
Sounds good and just piggyback on what NetworkTransform does for ClientAuth for now

text/0035-physics.md
```diff
@@ -0,0 +1,175 @@
...
+1. Prediction: velocity allows for a more accurate prediction.
+2. Extrapolation in case of packet loss.
+
+We do not support prediction currently so 1. is not relevant. Regarding extrapolation while this might improve the smoothness of the Rigidbody when extrapolating in case of a lag due to packet loss it is non trivial to implement this without resulting in jitter. What we would need to do here is transition the Rigidbody from a simply replicating mode into a dynamic Rigidbody again and run the physics simulation to extrapolate. We currently do not have a good way to make this transition smooth because we are not in control of the FixedUpdate and the physics simulation. Instead simply using the NetworkTransforms extrapolation mode results in a more smooth experience and keeps our implementation simple.
+
+## Why we do not synchronize triggers and collision events
+
+1. Collision and Trigger events contain information in reference types (classes) which is not something we could properly serialize. In addition these events contain information about the components and GameObject involved in the collision. We currently have no way to serialize that information over the network.
```
SamuelBellomo, line 116: could we have restrictions similar to parenting and make sure we sync at least the collisions with NetworkObjects? We'd be able to sync a net obj's ID.
I'm guessing it'd still be hard to get an ID from a material for example. Wouldn't it still be ok to sync partial information? It'd be better than nothing, no?

text/0035-physics.md
```diff
@@ -0,0 +1,175 @@
...
+
+2. Assuming we could serialize the data of the collision event there's an other issue. When sending collisions we could either send them reliable or unreliable both of them have downsides:
+    1. Unreliable: Would kind of defeat the purpose because collisions and triggers would still not be fully reliable
+    2. Reliable: If the packet gets dropped the collision even will be delayed compared to the position sync.
```
SamuelBellomo, line 120: what would be the result? We'd see a ball bump (or miss the wall because of interpolation) on a wall, then 1+ RTT later we'd see the trigger event? Would it be that bad?

text/0035-physics.md
```diff
@@ -0,0 +1,175 @@
...
+}
+```
+*note the above works for a server authoritative use case only* <br>
+*This example does not synchronize any information about the collision. The Collision class is not network serializable and can't be sent over an RPC. If the `other` object of the Collision is a NetworkObject the id of the object could be sent in the RPC to the clients to reference the right object*
+
+## 3D and 2D Physics
+
+3D and 2d Physics work exactly the same. The only difference is that the Rigidbody2D and NetworkRigidbody2D components need to be added to the GameObject instead of the Rigidbody and NetworkRigidbody components. The `Update` simulation mode of Physics2D is supported.
+
+# Reference-level explanation
```
SamuelBellomo, line 80: might be worth putting a note for Brian when he reads this, that a lot of the bellow should also be included

text/0035-physics.md
```diff
@@ -0,0 +1,175 @@
...
+
+Because of the two reasons above we will not do collision and trigger synchronization above and instead ask people to write their own using reliable RPCs if they need a collision event to be reliably synced to all clients. 
+
+## Switching Kinematic Mode at Runtime
+
+Sometimes for gameplay reasons switching the kinematic mode of the Rigidbody is needed. We do support switching the kinematic mode of the Rigidbody on the controller of the Rigidbody. The kinematic mode of the Rigidbody will be synchronized with a NetworkVariable to all other peers. This allows switching the owner of the Rigidbody 
+
+## Synchronizing Rigidbody Parameters
+
+We will not synchronize any Rigidbody parameters besides `IsKinematic`. For instance we will not synchronize the Mass or the Drag or the Constraints of a Rigidbody. The reasons we do not synchronize them are to save bandwidth and to keep the use cases of Rigidbody flexible. If synchronization of a specific Rigidbody parameter is needed users can add this with a custom script using a NetworkVariable.
```
SamuelBellomo, line 130: "to save bandwidth" --> maybe I'm missing something, that's a few bytes at spawn, then nothing if it doesn't change, no?
Not sure I understand the flexibility bit.

text/0035-physics.md
LukeStampfli: We might be able to do something but it would be very limited. The collision event contains information such as:
- The collider it collided with (we cannot synchronize this it's not a NetworkBehaviour and one Rigidbody can have many and they can be on child objects.
- Contacts, contact points, velocities, impulse could in theory be synced but they would not reflect the situation on the client at the time the `NetworkCollision` would be fired.

text/0035-physics.md
LukeStampfli: People might change these values but not change velocities. For instance it is quite common to adjust the mass or add constraints to a RigidBody at runtime. If the `controller` of the body never changes there is no need to synchronize those values.

text/0035-physics.md
LukeStampfli: I think delayed triggers/collisions is something we could live with. But it's just one more thing which makes syncing collision events iffy.

Since there are so many nuances and different ways to do it right I'd rather tell people to implement a simplified version which fits their need instead of trying to provide a one size fits all solution.

text/0035-physics.md
SamuelBellomo: we could still have a "DelayedCollisionPoint", telling users this isn't an exact point, but still something that happened in the past if you want to play some sparkle VFX for example?
It's not the mountain I want to die on, if you don't think that'd add value, we can close this :)

text/0035-physics.md
SamuelBellomo: that's a dirty check though no? That's different from saying we'll **never** sync these values. It's a bit like what Fatih did with NetworkTransform's state and per-axis synchronisation.

